PR #21, opened 2022-03-31
Changing bitmask-core to adapt to the new v3 manifest

josediegorobles:
Because of the difficoulties of passing messages for get data from local storage, i change bitmask-core to get and return the necessary data so the setting and getting will make all in the javascript part.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,4 @@
 BITCOIN_EXPLORER_API_MAINNET=https://blockstream.info/api
 BITCOIN_EXPLORER_API_TESTNET=https://blockstream.info/testnet/api
 BITCOIN_EXPLORER_API_SIGNET=https://blockstream.info/signet/api
-NODE_SERVER_BASE_URL=https://it62wjo9h3.execute-api.us-east-2.amazonaws.com/
+NODE_SERVER_BASE_URL=https://fa468o0r35.execute-api.us-east-2.amazonaws.com/prod/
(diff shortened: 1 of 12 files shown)


cryptoquick:
Good thinking! I definitely agree, persistence should be left up to the user. In our case, it's either a chrome extension or a web page. In v3, there's the sandbox, which might not be terribly persistent or have other limitations.

BTW, check this out in CI:
https://github.com/diba-io/bitmask-core/runs/5769903437?check_suite_focus=true

josediegorobles:
> Good thinking! I definitely agree, persistence should be left up to the user. In our case, it's either a chrome extension or a web page. In v3, there's the sandbox, which might not be terribly persistent or have other limitations.
> 
> BTW, check this out in CI: https://github.com/diba-io/bitmask-core/runs/5769903437?check_suite_focus=true

Yeah, the tests are failing now, i add some paremeters to the methods and now some of them are returning other types; we must modify the tests for that, i think is better a work of mine because i've been modifying the library.

josediegorobles:
@cryptoquick could you review the tests(and the code)? They are compiling but the wallet don't have any balance so one of the tests is failing.

cryptoquick:
Sounds good, @josediegorobles, will do.

cryptoquick (approved):
Code looks good, and it passes tests. :+1:

PR #22, opened 2022-04-15
Test fixes

cryptoquick:
It looks like there was a payment made to the original wallet that broke tests. Not sure how that happened. I also added .env to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /target
 /pkg
 .DS_Store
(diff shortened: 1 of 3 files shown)


cryptoquick:
Please note, we shouldn't add the .env to the gitignore, since it's needed to deploy to npm.

josediegorobles (approved):
Ok, all is working

PR #36, opened 2022-04-25
Changing how new address are generated and the transfer process

josediegorobles:
Adding /forget and /enclose calls.


cryptoquick (approved):
The code looks good, but still pending issues noted in the extension

PR #44, opened 2022-06-02
Blockstream to mempool.space

josediegorobles:


cryptoquick (approved):
Seems like a simple enough change! Try to open the sandbox iframe in the network inspector to verify it's using the proper API. I'm really curious if this will fix our Esplora reliability issues.

PR #46, opened 2022-06-15
Adding Error: to returns of wasm functions

josediegorobles:


cryptoquick (approved):
Actually, there might be an issue

src/lib.rs
```diff
@@ -392,7 +392,7 @@ pub fn send_sats(
         .await;
         match transaction {
             Ok(transaction) => Ok(JsValue::from_string(transaction)),
-            Err(e) => Ok(JsValue::from_string(format!("{} ", e))),
+            Err(e) => Ok(JsValue::from_string(format!("Error: {} ", e))),
```
cryptoquick: Shouldn't this be returning an Err? That way the async/await calls in JS can throw properly.

cryptoquick (changes requested):

src/lib.rs
josediegorobles: that's probably the best, i think that i go this route in the moment because some problem doing that and i take the fast route.
I will try later making that with errors.

cryptoquick: :+1:

cryptoquick (approved):
Approved!

PR #50, opened 2022-06-23
Refactor node URL generation out into its own function.

cryptoquick:


josediegorobles (approved):
ACK to the code, is much better and cleaner now, i will merge as soon as is mannually tested by me

Issue #45, opened 2022-06-09
Rename `get_pubkeyhash` to `get_xpub_hash`


cryptoquick:
It's confusing, since a pubkey is derived from an extended public key, and isn't unique to a seed + derivation path, which can't be considered a form of a unique identifier. But the actual code is indeed using the xpub, so it's misnamed.

Also, the weakly-typed tuple return type for `get_mnemonic` should be a strongly typed named struct:
https://github.com/diba-io/bitmask-core/blob/1c72f37f4d76b70d0a833f3eba24647c8fb7ca8f/src/operations/bitcoin/secret.rs#L39

cryptoquick:
Through the numerous refactors since this issued was filed, I'd say it's safe to close this.

PR #53, opened 2022-06-29
Platform Isomorphicity

### LINKED ISSUE #9: Isomorphic Compilation
Right now, the module only builds for WASM, but it would be very nice to have this functionality be capable of being built for native Rust.

- [ ] Isomorphic support from both bitmask-core and BDK
    - Issue tracking an obstacle to this: https://github.com/bitcoindevkit/bdk/issues/165#issuecomment-1061283739
    - Can get by using the `bdk-macros` library in the meantime
- [ ] Separate web and native libs using conditional compilation

cryptoquick:
Closes #9 #45


cryptoquick:
Screenshot from 2022-06-28 20-35-44
@josediegorobles :point_up:

PR #54, opened 2022-06-30
Adding import_accept

josediegorobles:


josediegorobles:
Adding import_accept to lib.rs and imprt_accept and accept_transaction to web.rs, now compiles but have the bdk 0.19 problem yet

cryptoquick (approved):
ACK, looks good to me.

PR #55, opened 2022-07-01
Fix linter error.

cryptoquick:
I think the issue is, no type is provided to serde to deserialize into, so I just passed the response string directly into the logger.


josediegorobles (approved):
ACK

PR #56, opened 2022-07-02
Embedded RGB

cryptoquick:
- [x] issue asset (since this would need to replace rgb-node)
- [x] import asset by genesis (importing by rgb id will be separate work)
- [x] creating blinded utxos to receive assets
- [x] initial work for initiating a transfer
- [x] debugging and testing asset transfer
- [ ] signing and sending asset transfer PSBT
- [ ] accepting asset consignments
- [ ] load asset by consignment

Note: I had to upgrade from BDK 0.18 because both BDK and RGB rely on rust-bitcoin 0.28.1, which would otherwise cause trait conflicts.

- Updated to BDK 0.20
- Add Issue Asset
  - We should be able to do asset genesis stuff now
- Changed all the descriptors into string references. That shouldn't mess with wasm-pack too much, and that streamlines our code (minimizes clones).
- `fund_wallet` (fund vault) should return a struct with explicitly-labeled txids + vout. Please verify I did this correctly, @Jose D Robles 
- Better logging in asset test
- Add `get_network` method
- Return actual signed tx in `sign_psbt`


josediegorobles:
The first thing is that if we don't have inmediate perspectives to integrate all of this with wasm, we must retain two vias for doing every thing, the old way with diba-rgb-node (that is going to be update with bitmask-node and be based in the same bitmask-core method) and the new way.
So for start I suggest to move old blind code to web.rs (with a comment about that) and the new to stay in lib.rs.

src/lib.rs
```diff
@@ -436,7 +473,7 @@ pub async fn import_accept(
         Ok(_accept) => {
             let wallet = get_wallet(rgb_tokens_descriptor, None).await;
             let unspent = wallet.as_ref().unwrap().list_unspent().unwrap_or_default();
-            let asset = get_asset(Some(asset), None, unspent, node_url).await;
+            let asset = get_asset_by_contract_id(Some(asset), None, unspent, node_url).await;
```
josediegorobles: We need three arguments here, maybe asset.id, unspent, node_url

src/operations/rgb/import_asset.rs
```diff
@@ -12,41 +10,21 @@ use crate::{
...
-    last_name: String,
-}
+    todo!();
 
-#[derive(Debug, Serialize, Deserialize)]
-struct PersonResponse {
-    data: String,
-    method: String,
-    headers: HashMap<String, String>,
+    Ok(assets)
```
josediegorobles: Not compiling because of this, maybe it can returns provissionally an Error?

cryptoquick:
@josediegorobles Just finished `get_asset_by_genesis`, check it out:

Screenshot from 2022-07-15 14-24-18

If someday we get the extension compiling to wasm, this method should work completely clientside. `get_asset_by_id` will actually have to be separate work, since it requires a persistence layer (either storm or s3).

josediegorobles:
> @josediegorobles Just finished `get_asset_by_genesis`, check it out:
> 
> Screenshot from 2022-07-15 14-24-18
> 
> If someday we get the extension compiling to wasm, this method should work completely clientside. `get_asset_by_id` will actually have to be separate work, since it requires a persistence layer (either storm or s3).

I have to check it out.

cryptoquick:
Still debugging this, I'm very close. BTW, the best way to test this is to run:

`RUST_LOG=bitmask_core=debug TEST_WALLET_SEED="..." cargo test`

josediegorobles:
Hi I've been testing with different seed words and is always with that:
---- asset_import stdout ----
thread 'asset_import' panicked at 'List of transactions is not empty', tests/asset.rs:45:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

.env
```diff
@@ -1,4 +1,7 @@
 BITCOIN_EXPLORER_API_MAINNET=https://mempool.space/api
 BITCOIN_EXPLORER_API_TESTNET=https://mempool.space/testnet/api
 BITCOIN_EXPLORER_API_SIGNET=https://mempool.space/signet/api
+BITCOIN_ELECTRUM_API_MAINNET=pandora.network:50001
+BITCOIN_ELECTRUM_API_TESTNET=pandora.network:60001
+BITCOIN_ELECTRUM_API_SIGNET=pandora.network:60601
```
josediegorobles: But it's working for you like this?

src/operations/rgb/descriptor_wallet.rs
```diff
@@ -0,0 +1,66 @@
+/*
```
josediegorobles, line 1: What is the purposed of this? Is al commented out

src/util.rs
```diff
@@ -4,16 +4,60 @@ use gloo_net::http::Request;
...
+        let output = vec![$(String::from($arg.to_owned()),)+].join(" ");
         #[cfg(target_arch = "wasm32")]
-        gloo_console::log!(format!("{}", output));
+        gloo_console::info!(format!("{}", output));
         #[cfg(not(target_arch = "wasm32"))]
         log::info!("{}", output);
     };
 }
 
+#[macro_export]
```
josediegorobles, line 17: I love this

josediegorobles (commented):
Great work for now. The part of making it call endpoints when is in wasm is not done, is it? If it's not done I will do it while completing the lambdas

josediegorobles:
Hi, about send_tokens we need to do some refactoring.

I'm trying to use a lambda with this. For this to not duplicating code, we need that transfer_asset be like this:

// Code (executable in wasm) that get the psbt (without rgb info), allocations, inputs.
// Calling another method rgb-only if we aren't in wasm and the lambda if we are in wasm. The method must be call with previous psbt, allocations, inputs and other info, contract_id, amount etc and returns the psbt with the rgb-info and the consignment, disclosure etc
// Code (executable in wasm) that finalize and signs and check if it's correctly send the PSBT of the previous line.

Do you think could be factible this refactoring, or maybe there are a better approach?

josediegorobles:
> Hi, about send_tokens we need to do some refactoring.
> 
> I'm trying to use a lambda with this. For this to not duplicating code, we need that transfer_asset be like this:
> 
> // Code (executable in wasm) that get the psbt (without rgb info), allocations, inputs. // Calling another method rgb-only if we aren't in wasm and the lambda if we are in wasm. The method must be call with previous psbt, allocations, inputs and other info, contract_id, amount etc and returns the psbt with the rgb-info and the consignment, disclosure etc // Code (executable in wasm) that finalize and signs and check if it's correctly send the PSBT of the previous line.
> 
> Do you think could be factible this refactoring, or maybe there are a better approach?

Here

https://github.com/diba-io/bitmask-core/pull/64/files

I create two unimplemented methods, one in lib, rgb_tweak, that is going to be called from the bit mask-extension. The other, rgb_tweakd is going to be called from transfer_assets (and probably with a different signature).

josediegorobles:
> > Hi, about send_tokens we need to do some refactoring.
> > I'm trying to use a lambda with this. For this to not duplicating code, we need that transfer_asset be like this:
> > // Code (executable in wasm) that get the psbt (without rgb info), allocations, inputs. // Calling another method rgb-only if we aren't in wasm and the lambda if we are in wasm. The method must be call with previous psbt, allocations, inputs and other info, contract_id, amount etc and returns the psbt with the rgb-info and the consignment, disclosure etc // Code (executable in wasm) that finalize and signs and check if it's correctly send the PSBT of the previous line.
> > Do you think could be factible this refactoring, or maybe there are a better approach?
> 
> Here
> 
> https://github.com/diba-io/bitmask-core/pull/64/files
> 
> I create two unimplemented methods, one in lib, rgb_tweak, that is going to be called from the bit mask-extension. The other, rgb_tweakd is going to be called from transfer_assets (and probably with a different signature).

My idea is only to extract the rgb code from there and put in another method, the wasm and lambda parts are going to be my thing.

cryptoquick:
@josediegorobles Feel free to move stuff around, just do me a favor, in the refactor, try to maximize code reuse as you move it into separate functions.

josediegorobles:
> @josediegorobles Feel free to move stuff around, just do me a favor, in the refactor, try to maximize code reuse as you move it into separate functions.

Yeah that's the idea.

josediegorobles (approved):

josediegorobles:
The tests are passing! I have to review the code.

josediegorobles (approved):

josediegorobles:
It's ok as far as I can see it. Some thing are going to be adapted to have a wasm compile, I'm working in that in https://github.com/diba-io/bitmask-core/pull/64 so is not part of this review

josediegorobles (approved):

src/operations/rgb/descriptor_wallet.rs
cryptoquick: It's in case I needed to embed the descriptor wallet, and it seems like I will need to.

src/util.rs
cryptoquick: Haha, thanks, yes, it helps a lot

.env
cryptoquick: I changed it to use ours, but on the server, please configure it to run with a `esplora.bitmask.app` hostname.

josediegorobles (approved):

josediegorobles (approved):

PR #58, opened 2022-07-28
Publishing data

josediegorobles:
I need to access to bitmask_core::data::structs::SealCoins for bitmask_lambda. It's probable that we will need to access to other structs. Provissionally I added a pub to mod data.., but if you think of a better way, @cryptoquick,  we can do it.


cryptoquick (approved):
Great, that works! If you need anything else, feel free to create more PRs.

PR #59, opened 2022-07-29
Adding get_asset

josediegorobles:
For the lambdas we need a without node_url and descriptor version of the method import_asset

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -272,39 +272,22 @@ pub fn create_asset(
     })
 }
 
-pub async fn get_asset(contract_id: Option<&str>, genesis: Option<&str>) -> Result<ThinAsset> {
-    match genesis {
-        Some(genesis) => {
-            info!("Getting asset by genesis:", genesis);
-            get_asset_by_genesis(genesis)
-        }
-        None => match contract_id {
-            Some(contract_id) => {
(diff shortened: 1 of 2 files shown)


cryptoquick:
This is just the same code as here, right?

https://github.com/diba-io/bitmask-core/blob/6a080391e8b3c629202d1f662784b9367383a095/src/lib.rs#L242-L268

Keep in mind, importing via contract_id might not be a thing we can support until we have a storage solution, like maybe a modified storm node.

josediegorobles:
> This is just the same code as here, right?
> 
> https://github.com/diba-io/bitmask-core/blob/6a080391e8b3c629202d1f662784b9367383a095/src/lib.rs#L242-L268
> 
> Keep in mind, importing via contract_id might not be a thing we can support until we have a storage solution, like maybe a modified storm node.

Yeah, I'm going to do it other way.

cryptoquick:
Okay, this can be closed, then?

josediegorobles:
> Okay, this can be closed, then?

Yeah I wanted to make sure the method can be called without using bdk, for the lambdas/server.

I made optional the descriptor and that's all.

I will make a PR for transfers doing the same.

cryptoquick:
Okay, I'll close it, then.

PR #61, opened 2022-07-30
Rgb core ht get asset and transfers

josediegorobles:


cryptoquick (approved):
Good improvement, thanks!

PR #63, opened 2022-07-31
Add BDK sqlite wallet database for desktop

cryptoquick:
The BDK sled implementation doesn't work well since it doesn't actively flush to disk. The sqlite implementation seems to work fine, though.

This feature is actually a LOT slower than the memorydatabase, and doesn't seem to cache anything at all, so I'm opening this up only as a draft pull request.


josediegorobles:
Ok, I understand, I don't have problems with the code. And the separation between wasm32 and not wasm32 targets must be part of the RGB integration PR.

josediegorobles (approved):

PR #64, opened 2022-08-01
Rgb core ht with modified transfer

josediegorobles:


josediegorobles:
It's building and passing the tests, but the method in lib.rs is not fully functional yet.

PR #65, opened 2022-08-04
Refactor get_wallet await performance.

cryptoquick:
Test runs 2x faster.


josediegorobles:
Ok, don't see any problem with the code, but is not building bit mask-extension with this.

Needs to use wasm-pack with nightly channel, how we configure the extension for that?

josediegorobles:
It's compiling now.

josediegorobles (approved):
Tested ACK

PR #67, opened 2022-08-04
Refactor get_wallet await performance.

cryptoquick:
Runs roughly x2 faster.


josediegorobles (approved):

josediegorobles (approved):

josediegorobles (approved):

cryptoquick:
It's very likely we'll get BDK to compile to WASM soon and if so, we won't need this code.

PR #68, opened 2022-08-05
Refactor-await-pef-restoring-env

josediegorobles:
Chcking is all well with old envs: if it's ok I think we can pass that because we know in the other PR is ok with the new envs..
By other side, I think this env must be hide.


josediegorobles:
I don't understand this, the tests are being passed locally

cryptoquick:
Maybe this is relevant:
https://discord.com/channels/753336465005608961/978744259693916230/1004431091144347749

If so, we may be held up by a WASM release of the latest BDK.

It looks like the rust-bitcoin folks are really close:
https://github.com/rust-bitcoin/rust-bitcoin/pull/1109

josediegorobles:
It can be that, yeah

PR #70, opened 2022-08-27
Embedded RGB fixes

cryptoquick:
Includes API changes from earlier this week, plus solves some issues in asset transfer, and includes latest changes from both of Zoe's PRs in rgb-node:
- https://github.com/RGB-WG/rgb-node/pull/196
- https://github.com/RGB-WG/rgb-node/pull/195

To be clear, the code duplication arrangement is temporary until we can build & contribute Embedded Storm upstream (zmq inproc / in-memory btreemap instead of sled).

The code here only fails because BDK fails to sign the PSBT because it fails to satisfy some condition. It's possible we'll need to use the RGB descriptor wallet to sign and broadcast the PSBT.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,9 @@ pub mod web;
 
 use data::{
     constants,
-    structs::{AssetResponse, SatsInvoice, ThinAsset, TransferResponse, VaultData},
+    structs::{
+        AssetResponse, FundVaultDetails, SatsInvoice, ThinAsset, TransferResponse, VaultData,
+    },
 };
 
 use operations::{
@@ -222,6 +224,7 @@ pub fn set_blinded_utxo(utxo_string: &str) -> Result<BlindingUtxo> {
         txid: Txid::from_str(split.next().unwrap())?,
         vout: split.next().unwrap().to_string().parse::<u32>()?,
     };
+
     let (blind, utxo) = blind_utxo(utxo)?;
 
     let blinding_utxo = BlindingUtxo {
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -246,33 +202,9 @@ pub fn create_asset(
...
-                info!(format!("asset: {asset:?}"));
-                match asset {
-                    Ok(asset) => Ok(asset),
-                    Err(e) => Err(format_err!("Server error: {e}")),
-                }
-            }
-            _ => Err(format_err!("Error: Unknown error in import_asset")),
-        },
-    }
+pub fn import_asset(genesis: &str) -> Result<ThinAsset> {
```
josediegorobles, line 205: Ok, the intention of this method is that if we have the descriptor , it returns how much tokens you have.
But really what we need for lambda is that: that if you pass the genesis or contract id AND utxo/blinded you get the ThinAsset with how many tokens you have. And if there are not utxo/blinded, then we simple get what it's being done now.

So instead of rgb_tokens_descripr, must be, utxo: Option<&str>, blinded: Option<&str> and the rest the same

I think the better is that I do a PR with my propose, but bettr give me your opinion on that.

josediegorobles (approved):

src/lib.rs
cryptoquick: This method can only get the asset by full genesis, but the genesis should be stored via lambda keyed by rgb id, and retrieved so it can be passed to this method by the client. They should be two separate operations, so that the contents can be encrypted end-to-end. Eventually all RGB methods will be run in WASM or clientside, the RGB lambdas are temporary, but the storage lambdas are permanent.

We should have a separate method that processes an entire consignment and returns all assets and balances. I haven't worked on that yet, but will do after I get transfers working.

Does that all make sense / sound good?

josediegorobles: It makes sense, of course.

cryptoquick: Sounds good, I'll merge now

PR #72, opened 2022-08-27
Rgb core ht api changes with modified transfer

josediegorobles:


cryptoquick:
I'll admit, I actually don't understand what key tweaking is for, or where it's used in RGB...

josediegorobles:
Sorry, i had to explain this:

The idea was follow same path that previous bitmask-extension:

1) Method in web.rs that is going to be executed in the extension/web where BDK forms a base PSBT ready for sign but without information about rgb assets

2) Method in lib that lambda is going to use for rgb-izing the PSBT: that was what the rgb-node does and now must be this rgb_tweaking method

3) Method in web.rs to sign the psbt that the lambda returns and send to the network.

We must have 1 and 3 in lib.rs too, so current send_assets calls 1,2,3 consecutively with essentially the same code and results that current send_assets.

I'm explaining well?

src/operations/rgb/send_tokens.rs
```diff
@@ -890,16 +967,9 @@ pub async fn transfer_asset(
     // btc-hot sign ${PSBT} ${DIR}/testnet
     // btc-cold finalize --publish testnet ${PSBT}
     let tx = sign_psbt(assets_wallet, psbt.into()).await?;
+    // let tx = sign_psbt(full_wallet, psbt.into()).await?;
 
-    let txid = tx.txid().to_string();
```
cryptoquick, line 894: This line got lost

src/operations/rgb/send_tokens.rs
```diff
@@ -735,13 +726,99 @@ pub async fn transfer_asset(
...
+        Err(_err) => return Err(anyhow!("Error making transition")), // TODO: make a more informative error compatible with the method
+    };
+    let (mut consignment, consignment_details) =
+        compose_consignment(&contract, outpoint_for_consignment.clone()).await?;
     psbt.set_rgb_contract(contract)?;
     debug!("RGB contract successfully set on PSBT");
-
-    // Embed information about the state transition into the PSBT
-    // rgb-cli -n testnet transfer combine ${CONTRACT_ID} ${TRANSITION} ${PSBT} ${UTXO_SRC}
-    // rgb-node -> cli/command -> TransferCommand::Combine
```
cryptoquick, line 744: Why were these comments removed?

cryptoquick (changes requested):
I still can't say I understand what this is for or why things are moved around. Can you explain more about what this supposed to do?

Also, can you add to the test for this? Maybe seeing it used will give me a better idea.

josediegorobles:
Yeah, this pr needs a good update

josediegorobles:
@cryptoquick I think I start with a too premature version this PR and is too messy, so I'm gonna start again is gonna be better and faster. I create another PR: https://github.com/diba-io/bitmask-core/pull/83 I hope that I can clarify intention and make a good start before I have to cut work, anyway we can discuss there the PR.

PR #74, opened 2022-09-16
WASM fixes

cryptoquick:
Pushed BDK back to 0.19 for WASM builds only.
Fixed up web.rs files to reflect new changes.
Started some work on Lambda calls, but that's not finished yet.
Bumped the version in preparation for release.
Also, removed `import_accept`, I don't think we need that.


josediegorobles (approved):

cryptoquick:
@josediegorobles Just saw you reviewed this actually, ignore my review request, merging.

PR #75, opened 2022-09-20
Fund vault fix

cryptoquick:
This is just a band-aid, I'll make an issue for parameterizing the function with a specific amount of coins

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -290,11 +290,11 @@ pub async fn fund_wallet(
 
     let asset_invoice = SatsInvoice {
         address: address.unwrap(),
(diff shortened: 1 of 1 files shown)


josediegorobles (approved):
Ok, better to have no problems with that but we will need parametrize this for mainnet and have a lower number of sats, it's a lot of sats in the end.

grunch (approved):
Tested ACK.

It works but I'm still not sure why we need to create assets and uda invoices with 10000 sats

cryptoquick:
> It works but I'm still not sure why we need to create assets and uda invoices with 10000 sats

It's to pay for fees

PR #79, opened 2022-09-20
RGB Transfer fixes

cryptoquick:


grunch (approved):
LGTM!

cryptoquick:
lol, thanks guys! I still want to try to get the test repeatable, so there's still a couple of things to be done there. But that work is actually encapsulated in #81, so we can merge this work for now, since it also fixes the wasm lint, and that'll be useful in other branches. Also, BDK merged the branch I'm using, but we can just switch to 0.23 once that's released, so this work is fine for now.

PR #80, opened 2022-09-20
Manually set fees

### LINKED ISSUE #16: Manually set fees
There should be a way to manually provide a fee rate to each on-chain method.

grunch:
Fix #16

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,7 @@ extern crate amplify;
 use std::str::FromStr;
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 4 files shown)


cryptoquick (changes requested):
Looks good, and good thinking to change the other send sats methods, but can you double-check to see if this passes `cargo fmt` / `cargo clippy` ?

grunch:
> Looks good, and good thinking to change the other send sats methods, but can you double-check to see if this passes `cargo fmt` / `cargo clippy` ?

Done

cryptoquick (approved):
Great, LGTM

PR #82, opened 2022-09-21
Parameterize fund wallet with sats amount

### LINKED ISSUE #76: Parameterize `fund_wallet` with a specific number of sats
- [ ] We should have two separate values for the assets vault and UDAs vault
- [ ] In the test, default it to `10_000` for both, I recommend funding testing wallets with about 100k sats initially

grunch:
Fix #76


cryptoquick (approved):
Looks good!

PR #83, opened 2022-09-22
Clientside Transfer PSBT construction

josediegorobles:
I restart this PR for clarify intention and because the previous start with premature versions of transfers and stated to be a little (or a lot really) messy

### DIFF
--- a/src/data/structs.rs
+++ b/src/data/structs.rs
@@ -132,6 +132,14 @@ pub struct TransferResponse {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TransferLambdaResponse {
+    pub consignment: String,
+    pub disclosure: String,
+    pub txid: String,
+    pub witness: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct ValidateRequest {
     pub consignment: String,
 }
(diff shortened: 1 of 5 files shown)


josediegorobles:
Thing is that:

In web.rs send_tokens can't simply call a lambda sending all the info... because we don't want to send the descriptors obviously.

The ideal thing would be to do the same as the process with rgb-node is:

1) First compose the PSBT with the info about the bitcoin transfer, with the BDK library (this method is going to be executed equal in wasm and in native).
2) Then "tweak" the PSBT for add the tokens transfer info. Only in native because this step requires rgb libraries and can't be executed in wasm (for now).
3) Sign and send the PSBT. Only with BDK, and can be executed in native and in wasm.

In native send_tokens is going to execute 1), then 2) then 3) consecutively, so we are going to split the code for doing that, leaving to do exactly the same.
In wasm we are going to execute 1, then send PSBT to a lambda that is going to execute 2 and with the answer we are going to execute 3.

1 --> compose_btc_psbt
2 --> compose_rgb_psbt
3--> that is sign_psbt

cryptoquick:
Oh, that's awesome!! If you can accomplish that, that'd be wonderful!! That's a good explanation, too, and was something I overlooked when I was just trying to get transfers working at all, haha...

cryptoquick:
@josediegorobles Is this a better title?

PR #85, opened 2022-09-28
Debugging transfer reliability

cryptoquick:
This is a terrible kludge and doesn't even work, but it's useful for the purpose of illustrating the issue.

Frustratingly, the issue doesn't arise every time the test is run, but when it does fail, nothing I can think of seems to get it to succeed.

I'd estimate the issue arises about half the time the test is run.


cryptoquick:
This code was way off the mark, now that we know what the issue is. Closing.

PR #94, opened 2022-10-07
Add lightning support with lndHub

grunch:
This is WIP, not ready to merge

PR #98, opened 2022-10-14
Add lightning support

grunch:
Fix #94


grunch:
Now we are using this one https://github.com/diba-io/bitmask-core/pull/100

PR #107, opened 2022-11-03
Cleanup get_assets_vault utxos

cryptoquick:
I'm not sure if this change is desirable. It's less explicit, but it looks cleaner.


josediegorobles (approved):

grunch (approved):
idiomatic and nicer 🦀

PR #108, opened 2022-11-03
First testing wasm function

grunch:


src/web.rs
```diff
@@ -327,3 +327,17 @@ pub fn switch_host(host: String) -> Promise {
         Ok(JsValue::UNDEFINED)
     })
 }
+
+#[wasm_bindgen]
+pub fn ln_create_wallet() -> Promise {
+    set_panic_hook();
+
+    future_to_promise(async move {
+        match crate::create_wallet().await {
```
cryptoquick: let's do `crate::lightning::create_wallet`, and then export only `lightning` and not `lightning::create_wallet`. Then we can namespace the other methods later into their own modules.

grunch: I have it like that before this PR, I don't know why I changed 🥲

josediegorobles (approved):
LGTM

src/operations/lightning.rs
```diff
@@ -198,57 +202,55 @@ pub async fn auth(creds: Credentials) -> Result<Tokens> {
...
-    let response = post_json_auth(&url, &Some(req), Some(token)).await?;
+    let req = InvoiceReq { invoice };
+    let response = post_json_auth(&url, &Some(req), Some(&token)).await?;
     let response: PayInvoiceMessage = serde_json::from_str(&response)?;
 
     Ok(response)
 }
 
 /// Get successful lightning transactions user made. Order newest to oldest.
-pub async fn get_txs(token: &str, limit: u32, offset: u32) -> Result<Vec<Tx>> {
```
cryptoquick: I'd prefer to keep the internal methods as string references, and then pass them in by reference in web.rs. Be sure to see how it's done in the other web.rs methods.

cryptoquick (changes requested):
Good comments. I'd prefer to still use string references in methods used internally. Also, this isn't passing linters.

src/operations/lightning.rs
grunch: done the String -> &str, there is something that I don't get it, why linters are not passing, I see they are broken but in files that are not related with this PR

josediegorobles (approved):
ACK. All the clippy warnings are corrected in development, so is not necessary to crrect it here.

Only, you need to see this issue with tests:

error[E0603]: module `shared` is private
  --> tests/asset.rs:6:24
   |
6  | use bitmask_core::rgb::shared::Reveal;
   |                        ^^^^^^ private module
   |
note: the module `shared` is defined here
  --> /home/runner/work/bitmask-core/bitmask-core/src/operations/rgb.rs:10:1
   |
10 | mod shared;
   | ^^^^^^^^^^

src/operations/rgb.rs
```diff
@@ -7,7 +7,7 @@ mod import_asset;
 mod issue_asset;
 mod receive_tokens;
 mod send_tokens;
-pub mod shared;
+mod shared;
```
cryptoquick: Be sure to leave this like it is in `development`:
https://github.com/diba-io/bitmask-core/blob/a23e8efe3031a8cf6fad3fd507f4942a0d95a97e/src/operations/rgb.rs#L10

cryptoquick (changes requested):

josediegorobles (approved):

cryptoquick (approved):
LGTM!

PR #109, opened 2022-11-07
Adding check blinding factor to accept transfer

josediegorobles:


src/data/structs.rs
```diff
@@ -205,8 +206,7 @@ pub struct ValidateRequest {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct AcceptRequest {
     pub consignment: String,
-    // pub outpoint: OutPoint,
-    // pub blinding_factor: String,
+    pub reveal: Option<Reveal>,
```
cryptoquick: Let's make this a non-optional String (since accepting without a reveal doesn't make sense, right?)

Also, as a string, it can be parsed and encoded using these formatting and parsing methods:
https://github.com/diba-io/bitmask-core/pull/109/files#diff-d9c6a42fed07589e989e8da734d4ba5f8ce37f6cf4cd63b3a22906f6b9451d98R71

cryptoquick (changes requested):

src/data/structs.rs
josediegorobles: Good idea, much better like that. And really option make no sense

src/operations/bitcoin/secret.rs
```diff
@@ -32,11 +32,7 @@ fn get_descriptor<C: ScriptContext>(
     let origin: KeySource = (xprv.fingerprint(&secp), deriv_descriptor);
     let derived_xprv_desc_key: DescriptorKey<C> = derived_xprv.into_descriptor_key(
         Some(origin),
-        DerivationPath::default().child(ChildNumber::from_normal_idx(if is_change {
-            1
-        } else {
-            0
-        })?),
+        DerivationPath::default().child(ChildNumber::from_normal_idx(u32::from(is_change))?),
```
cryptoquick, line 35: Clever!

cryptoquick (approved):
Wow! This is all fantastic! Let's get this in!!

PR #113, opened 2022-11-20
Bump to RGB v0.8.1

oacdutra:
Update the RGB  to latest stable version and fix some compilation issues.

I working in other minor issues in RGB, but now is possible to work with release version.


cryptoquick (approved):
Looks great, thanks!

PR #116, opened 2022-11-25
Adding data for persistence

josediegorobles:
I need that when lambda calls core methods they returns more data about previous utxo, new (for the sender) utxo and about the blinded utxo, without changing the callings from bitmask extension.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -561,7 +561,7 @@ pub async fn send_assets(
     info!("Creating transfer PSBT...");
 
     #[cfg(not(target_arch = "wasm32"))]
-    let (consignment, psbt, disclosure, _, _) = transfer_assets(
+    let (consignment, psbt, disclosure, _change, _previous_utxo, _new_utxo) = transfer_assets(
         rgb_assets_descriptor_xpub,
         blinded_utxo,
         amount,
@@ -626,11 +626,12 @@ pub async fn transfer_assets(
     String, // json
     String,
     String,
+    String,
 )> {
     // use lnpbp::bech32::ToBech32String;
     use strict_encoding::strict_serialize;
 
-    let (consignment, psbt, disclosure, change, previous_utxo) = transfer_asset(
+    let (consignment, psbt, disclosure, change, previous_utxo, new_utxo) = transfer_asset(
         rgb_assets_descriptor_xpub,
         blinded_utxo,
         amount,
@@ -650,7 +651,14 @@ pub async fn transfer_assets(
     let change = serde_json::to_string(&change)?;
     let previous_utxo = serde_json::to_string(&previous_utxo)?;
 
-    Ok((consignment, psbt, disclosure, change, previous_utxo))
+    Ok((
+        consignment,
+        psbt,
+        disclosure,
+        change,
+        previous_utxo,
+        new_utxo,
+    ))
 }
 
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -561,7 +561,7 @@ pub async fn send_assets(
     info!("Creating transfer PSBT...");
 
     #[cfg(not(target_arch = "wasm32"))]
-    let (consignment, psbt, disclosure) = transfer_assets(
+    let (consignment, psbt, disclosure, _change, _previous_utxo, _new_utxo) = transfer_assets(
```
cryptoquick: Mind if we change these to a `TransferAssetsResponse` struct? Then we can just pass it around.

src/lib.rs
```diff
@@ -669,12 +682,14 @@ pub async fn accept_transfer(
     blinding_factor: &str,
     outpoint: &str,
 ) -> Result<AcceptResponse> {
+    info!("ajaja");
```
cryptoquick: lol

src/web.rs
```diff
@@ -268,7 +268,7 @@ pub fn send_assets(
 #[wasm_bindgen]
 pub fn accept_transfer(consignment: String, blinding_factor: String, outpoint: String) -> Promise {
     set_panic_hook();
-
+    crate::info!("aqui estamos");
```
cryptoquick: Maybe set a more informative log message if you want to leave this in.

src/operations/rgb/send_tokens.rs
```diff
@@ -191,7 +194,7 @@ pub async fn transfer_asset(
     let change_output = change_outputs.get(0).unwrap();
     debug!(format!("Selected change output: {change_output:#?}"));
 
-    let change = change
+    let change: BTreeMap<bp::seals::txout::blind::RevealedSeal, u64> = change
```
cryptoquick: It'd be better to use `RevealedSeal` from the imports section.

cryptoquick (changes requested):
Just a couple of changes, but it looks good.

grunch (approved):
LGTM!

src/web.rs
josediegorobles: No, i left several debug logs that are unnecessary...

src/lib.rs
josediegorobles: Yeah much better

josediegorobles:
Now I'm calling a new endpoint, declare, that persist the data of the receptor and the change data of the sender. Now i'm sending all as soon as the transaction is sent.

src/lib.rs
```diff
@@ -624,11 +633,14 @@ pub async fn transfer_assets(
     String, // bech32m compressed sten consignment
     String, // base64 bitcoin encoded psbt
     String, // json
+    String,
```
grunch: It would be great to comment those new strings, change, previous_utxo and new_utxo

src/operations/rgb/send_tokens.rs
```diff
@@ -290,7 +286,7 @@ pub async fn transfer_asset(
                 "Error constructing PSBT from RGB Tokens Descriptor: {err}",
             ));
             return Err(anyhow!(
-                "Error constructing PSBT from RGB Tokens Descriptor"
+                "jaja: Error constructing PSBT from RGB Tokens Descriptor: {err}"
```
grunch: 🤣

grunch (changes requested):

src/operations/rgb/send_tokens.rs
josediegorobles: Haha.

grunch (approved):

cryptoquick (approved):
ACK. LGTM!

PR #119, opened 2022-12-04
Send PayJoin

DanGould:
Change interface to accept a bip21 PjUri as defined in bip78. Only use address from the bip21 Uri, still pass amount in as a separate argument to send_sats.

I think the design choices to consider are mostly around passing how we'd like to pass the bip21 from send_sats. Where is the web interface that uses this?

The other question is whether or not to finalize and broadcast the original_psbt in the case of an unresponsive pj endpoint.

This was my first time handling errors in rust with anyhow. Wow, much ease!

May I please have some testnet sats tb1pn4g2sjt0qw9t43lj28quhnajs064n8frzalqjqueuvcas4h6hxlsau59qd ? I can't find a faucet to pay a bitmask taproot address. so futuristic.

You can get a bip21 receiver for this payjoin sender at https://testnet.demo.btcpayserver.org/login to test on testnet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ anyhow = "1.0"
 base64 = { package = "base64-compat", version = "1.0" }
 bech32 = "0.9"
 bip39 = "1.0"
+bip78 = { git = "https://github.com/dangould/rust-payjoin", rev = "8d9b7d6", features = ["sender" ] }
 bitcoin = "0.28.1"
 bitcoin_hashes = "0.10.0"
 console_error_panic_hook = "0.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -364,6 +364,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bip21"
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -30,6 +30,7 @@ anyhow = "1.0"
 base64 = { package = "base64-compat", version = "1.0" }
 bech32 = "0.9"
 bip39 = "1.0"
+bip78 = { git = "https://github.com/dangould/rust-payjoin", rev = "8d9b7d6", features = ["sender" ] }
```
cryptoquick, line 33: We need this to be published in order to publish our crate.

cryptoquick (changes requested):
Wow, nice! Just one change requested. I'm also going to see if I can write an automated test for this.

Cargo.toml
DanGould: certainly

DanGould:
Were you able to smoke test a payjoin on testnet?

PR #123, opened 2022-12-19
Support RGB121

oacdutra:
**TODO**
- [ ] Support contract parameters
- [ ] Support file attachments


oacdutra:
@cryptoquick,

I think this PR can be closed, correct?

cryptoquick:
@crisdut I suppose so, we'll definitely need to revisit this, but another time, maybe 0.10. Thanks for putting in the thought, though!

PR #125, opened 2022-12-24
Support to multi-asset transfer and regtest

oacdutra:
**Description**
The PSBT file construct operation does not support pass tweak as parameter and this causes some issues:
- We cannot be uses tweaked pukeys. 
- We cannot spend **bitcoins** sent  to address with commitments. 

This PR fix that with a custom version of the `psbt::construct` (_descriptor-wallet_).

**Scenarios**
The last three modifications (bump rgb to v0.8.1, terminal derivation, and this PR) allow us to perform all tapret operations:

**1 - Transfer Assets**
We need two UTXOs (one with an anchored asset and the other to change assets). After transfer, the bitcoins are sent to tweaked output, and the asset change is anchored in change utxo.

**2 - Transfer Assets (change address)**
We need to pass the terminal derivation to use the asset sent to change UTXO. This operation work the same way of the first scenario.

**3 - Transfer Bitcoins (commitments outputs)**
To spend bitcoin sent to the tweaked output, we need to pass the commitment message (obtained in the return of the asset transfer, the 'message' field of the disclosure JSON) in the input descriptor used to generate the PSBT file.

**Workarounds**
I need create a operation to generate PSBT in bitmask core to support the **scenario 3**. 

Before finalizing, I would like to know if these changes are pertinent to the project and answer any questions.

**Break changes**
- Change send_assets to transfer_assets (wasm)
- Remove send_assets


src/bin/bitmaskd.rs
```diff
@@ -46,7 +46,7 @@ async fn import(Json(asset): Json<AssetRequest>) -> Result<impl IntoResponse, Ap
 
 #[axum_macros::debug_handler]
 async fn transfer(Json(transfer): Json<TransferRequest>) -> Result<impl IntoResponse, AppError> {
-    let (consignment, psbt, disclosure, _, previous_utxo, _) = transfer_assets(
```
cryptoquick, line 49: Please use destructuring assignment, similar to this:
https://github.com/diba-io/bitmask-core/blob/42dc3fad86c8a2b0a8d290950744fb10c07742b0/src/lib.rs#L570-L575

cryptoquick (changes requested):
Wow, this looks pretty good. Just going to ask for one stylistic change. Also, can you modify the asset test to test multiple transfers?

oacdutra:
Hi,

I finished the implementation to support multiple-asset transfers.

**4 - New way to transfer assets**
- One asset to one beneficiary
- One asset to many beneficiaries
- Many assets to one beneficiary 
- Many assets to many beneficiaries

To avoid break changes, I created new methods to help us test and approve this PR.

Also, I created a new endpoint in bitmaskd daemon to test it.

**How this works**

For now, I have created a new request to support this feature; see bellow:

```
{
    "descriptor_xpub": "tr(..)",                    // RGB Descriptor
    "transfers": [
        {
            "asset_utxo": {
                "outpoint": "74..17:1",             // Anchored UTXO
                "terminal_derivation": "/0/0",      // Descriptor Terminal Derivation
                "commitment": ""                    // TapTweak (Optional)
            },
            "asset_amount": 1000,                   // Total Asset Amount
            "change_utxo": "6e..eb5:1",             // Change UTXO for RGB
            "beneficiaries": [                      // List of Blinded Utxo + Amount
                "10@txob1..rt",
                "20@txob1..ld"
            ],
            "asset_contract": "rgbc1..65x"          // RGB Contract
        }
    ]
}
```

**Limitations**

Today the multiple asset UTXO needs to derive from the same descriptor.

I running on regtest, that's works.

cryptoquick (changes requested):
I'm not worried about breaking changes, be sure to replace the old functionality with the new functionality, because the old functionality was essentially broken. Please get rid of the "transfers" plural endpoint, and v2 stuff. I don't want to maintain code that's already broken, does that make sense?

oacdutra:
> I'm not worried about breaking changes, be sure to replace the old functionality with the new functionality, because the old functionality was essentially broken. Please get rid of the "transfers" plural endpoint, and v2 stuff. I don't want to maintain code that's already broken, does that make sense?

Sure, I maintained the old version to help us mapping the changes in other projects. So I will make a new commit to remove all the old code.

josediegorobles (approved):
Great work and changes, you have my ACK.

PR #126, opened 2022-12-27
Add xpubkh to EncryptedWalletData struct

cryptoquick:
This is to support username/password situations where something unique to each wallet is used as a login without revealing the keys themselves. As an example use-case, we're using it for the lndhubx login.


josediegorobles (approved):
LGTM

PR #127, opened 2022-12-29
Adding tapret spend support - DIBA

cryptoquick:
Closes #125 

In the future, @crisdut and @grunch, let's work from the diba-io origin for git, it makes testing CI stuff like this easier.

**Description**
The PSBT file construct operation does not support pass tweak as parameter and this causes some issues:
- We cannot be uses tweaked pukeys. 
- We cannot spend **bitcoins** sent  to address with commitments. 

This PR fix that with a custom version of the `psbt::construct` (_descriptor-wallet_).

**Scenarios**
The last three modifications (bump rgb to v0.8.1, terminal derivation, and this PR) allow us to perform all tapret operations:

**1 - Transfer Assets**
We need two UTXOs (one with an anchored asset and the other to change assets). After transfer, the bitcoins are sent to tweaked output, and the asset change is anchored in change utxo.

**2 - Transfer Assets (change address)**
We need to pass the terminal derivation to use the asset sent to change UTXO. This operation work the same way of the first scenario.

**3 - Transfer Bitcoins (commitments outputs)**
To spend bitcoin sent to the tweaked output, we need to pass the commitment message (obtained in the return of the asset transfer, the 'message' field of the disclosure JSON) in the input descriptor used to generate the PSBT file.

**Workarounds**
I need create a operation to generate PSBT in bitmask core to support the **scenario 3**. 

Before finalizing, I would like to know if these changes are pertinent to the project and answer any questions.

**Break changes**
- Change send_assets to transfer_assets (wasm)
- Remove send_assets


oacdutra:
I added the docker in github action!

I will open a issue to some ideas to decrease the github action time.

cryptoquick:
@josediegorobles  There are some small changes I'll need to make, but otherwise, this should be ready for you to review. I'll finish mine tomorrow.

josediegorobles (approved):
LGTM. We are having a so good end of the year!

grunch:
What I could understand looks great, great job @crisdut 🥳

cryptoquick:
Just did some minor cleanup. If this passes CI, I'll get it merged. Thanks for the review, and @crisdut, excellent job on this one dude. Not only did we get regest support, which will help a lot, but your send_transfers was incredible, and displayed and excellent understanding of how RGB works. We're all super grateful for your work, dude. Thank you so much. :pray:

PR #132, opened 2023-01-04
Return error for bad mnemonic

cryptoquick:
Also adds a test and replaces some other expects with `?` so errors bubble upwards.


josediegorobles (approved):
ACK

PR #136, opened 2023-01-10
Reverting import asset and adding wasm to transfer

josediegorobles:
By mistake, import asset lost sending blinding and is necessary to retrieve data from lambdas.
Plus I add wasm_bindgen to transfer_assets


oacdutra (approved):

oacdutra:
OMG, I'm sorry, I think I removed that.

cryptoquick (approved):
I noticed this, but I thought this was intentional. Good catch, @josediegorobles.

PR #137, opened 2023-01-12
Updating transfer response and others

### LINKED ISSUE #138: Blinded UTXOs in Asset Import
I ran into an issue with asset_import:
https://github.com/diba-io/bitmask-core/blob/4b686330d79e748b759b572a61fd84215a259a0d/src/lib.rs#L316

In practice this results in a request like this:
```
{
  "asset": "rgb19jkjhpsjvlmktzhdax8dpmhwx59eafkqhhy3znswwf40l5e0ppuq6damnl",
  "utxos": [
    "266d2190920eb02e89c6a0425c2ce2fe2bd19f4660245023c97213eaef942c19:1",
    "txob1v4hymp5sd5j8rnn4sl0rz0se7w94rus8vhwmh72s6sy8kg7es9eqn3ml4u"
  ]
}
```

This results in this error:
```
Something went wrong: OutPoint not in <txid>:<vout> format
```

I'm not sure the best way to handle this. I'm hoping @crisdut will know more.

josediegorobles:
Updates for bit mask lambda to use it

Closes #138


src/lib.rs
```diff
@@ -565,7 +584,20 @@ pub async fn transfer_assets(transfers: TransfersRequest) -> Result<TransfersSer
     if status != 200 {
         return Err(anyhow!("Error calling {endpoint}"));
     }
-    Ok(serde_json::from_str(&transfer_res)?)
+
+    let transfer_res: TransfersSerializeResponse = serde_json::from_str(&transfer_res)?;
+
+    let _declare = async {
```
grunch, line 611: Just out of curiosity, wouldn't be better to remove the `_declare` var? like this:

```rust
    async {
        let declare_endpoint = &get_endpoint("declare").await;
        let (_declare_res, status) = post_json(declare_endpoint, &transfer_res.declare).await?;
        if status != 200 {
            return Err(anyhow!("Error calling {declare_endpoint}"));
        }
        Ok(status)
    }
    .await?;
```

josediegorobles: Yeah

src/operations/rgb/send_tokens.rs
```diff
@@ -141,7 +145,17 @@ pub async fn transfer_asset(request: TransfersRequest) -> Result<TransfersRespon
...
+        let consignment_serialize =
+            strict_serialize(&consignment).expect("Consignment information must be valid");
+        let consignment_serialize = util::bech32m_zip_encode("rgbc", &consignment_serialize)
+            .expect("Strict encoded information must be a valid consignment");
         transfers.push((consignment, beneficiaries.into_keys().collect::<Vec<_>>()));
+        transaction_info.push((
+            contract.contract_id().to_string(),
+            changes_info,
+            seal_coins,
+            consignment_serialize.clone(),
```
oacdutra: I was confused on this part.

Do you need the original contract or the new state?

src/data/structs.rs
```diff
@@ -268,16 +274,17 @@ pub struct AcceptResponse {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct BlindedOutpoint {
+pub struct BlindedOrNotOutpoint {
```
oacdutra, line 288: I think is good we need some functions to the struct, like `is_blinded`, `is_revealed`, etc... 

What do you think?

oacdutra (approved):
I verified that the transfer tests broke after the modifications.

Please, can you update each test with code bellow?

```rust
    info!("Accept transfer");
    for transfer in resp.declare.transfers {
        for beneficiary in transfer.beneficiaries {
            if beneficiaries.contains_key(&beneficiary.outpoint) {
                let reveal = beneficiaries
                    .get(&beneficiary.outpoint)
                    .expect("Beneficiary not found in transition");
                let accept_details = accept_transfer(
                    &transfer.consignment,
                    &reveal.blinding,
                    &reveal.utxo.to_string(),
                )
                .await?;
                debug!("Accept response: {:#?}", &accept_details);
                assert_eq!(accept_details.id, issued_asset.asset_id, "RGB IDs match");
            }
        }
    }
```

src/data/structs.rs
josediegorobles: Good idea. And if you have a better name for the struct, it would be good too.

src/operations/rgb/send_tokens.rs
josediegorobles: I need to identify all assets of same origin, so I'm using it.

cryptoquick:
@josediegorobles Is this ready for review?

src/lib.rs
```diff
@@ -527,45 +527,75 @@ pub async fn get_assets_vault(
...
-                beneficiaries: seals.into_iter().map(|s| s.to_string()).collect(),
-            }
-        })
-        .collect();
+
+    let mut transfers = vec![];
+    let mut change_transfers = vec![];
+    for (index, asset_transfer_info) in resp.transaction_info.iter().enumerate() {
+        transfers.push(FinalizeTransfer {
+            consignment: asset_transfer_info.consignment.clone(),
```
oacdutra: Hi @josediegorobles,

I think this line has incorrect, because the `transaction_info` does not contains the transfer state, but consignment contract.

I don't known if you need the original consignment information in this part, but before the change, we return the transfer state.

Because of this, the tests broke when try check the accept transfer.

oacdutra (changes requested):

src/lib.rs
josediegorobles: It's is what is being passed: let consignment_serialize =
            strict_serialize(&consignment).expect("Consignment information must be valid");
        let consignment_serialize = util::bech32m_zip_encode("rgbc", &consignment_serialize)
            .expect("Strict encoded information must be a valid consignment");

It's a simple String at this point, so I don't see he problem

oacdutra (approved):

grunch (approved):
LGTM!

cryptoquick (approved):
ACK

PR #139, opened 2023-01-18
Clippy is now linting tests

cryptoquick:
These lints were found in your modifications to the assets test, @crisdut, can you double-check to see if everything makes sense here? Should I remove the commented sections?

### DIFF
--- a/tests/asset.rs
+++ b/tests/asset.rs
@@ -1,10 +1,9 @@
 #![cfg(not(target_arch = "wasm32"))]
 
 use core::time;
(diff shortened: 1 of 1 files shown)


oacdutra:
> These lints were found in your modifications to the assets test, @crisdut, can you double-check to see if everything makes sense here? Should I remove the commented sections?

Hi @cryptoquick,

I think the tests broke because the state transfer is missing. I will comment in @josediegorobles PR, ok?

cryptoquick:
Oh, gotcha! I completely misunderstood. Thanks for pointing that out. I'll close this.

PR #141, opened 2023-01-18
Expose state transfers and fix assets tests 

oacdutra:
I tried explain [here](https://github.com/diba-io/bitmask-core/pull/137#discussion_r1073893315) why we need expose the state_transfer, but I couldn't explain it clearly, sorry. We need this information to process  `accept_transfer` step correctly.

I performed the asset transfer tests, all tests works.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -527,6 +527,9 @@ pub async fn get_assets_vault(
 
 #[cfg(not(target_arch = "wasm32"))]
 pub async fn transfer_assets(transfers: TransfersRequest) -> Result<TransfersSerializeResponse> {
+    use rgb_std::{Contract, StateTransfer};
+    use strict_encoding::strict_serialize;
+
     use crate::data::structs::{BlindedOrNotOutpoint, ChangeTansfer, DeclareRequest};
 
     let resp = transfer_asset(transfers).await?;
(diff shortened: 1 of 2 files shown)


cryptoquick (approved):
ACK, LGTM... Though, can you explain to me what the difference is between a consignment and a state transfer? A state transfer is just a portion of the consignment, correct?

cryptoquick:
Nevermind, @crisdut, I just read your comment [here](https://github.com/diba-io/bitmask-core/pull/137#discussion_r1073893315), that explanation is sufficient, I think.

oacdutra:
> ACK, LGTM... Though, can you explain to me what the difference is between a consignment and a state transfer? A state transfer is just a portion of the consignment, correct?

Yes, Sure

There is no big difference between the two, just that the state transfer has the current state of the contract plus the state transition data. The `state transfer`  is the final result of the rgb-cli transfer finalize operation.

cryptoquick (changes requested):
Can you add a few comments that explains what just explained in the code for me, @crisdut ?

josediegorobles (approved):
LGTM

cryptoquick (approved):
Thanks for adding the comment, @crisdut! LGTM. Merging.

PR #142, opened 2023-01-18
Fix Tests Logs

oacdutra:


cryptoquick:
I'm not too worried about this, and it's helpful to have the log output for debugging, especially now that we're on regtest, so the test environment is different than on local.

PR #143, opened 2023-01-19
Add PSBT publish file test

oacdutra:
I needed to change some values ​​we test to avoid dust transactions.


josediegorobles (approved):
LGTM

PR #144, opened 2023-01-23
Multiple Transfers test case

cryptoquick:
I wanted to create a failing test to test multiple transfers. This will help as we evaluate different approaches to create new change UTXOs, and for the wallet to determine which is the change UTXO.

I don't think this should block our 0.5 release, however. We can add that functionality as a point release. For now, I've made the test pass, and added some comments so it's clear why we expect that particular one to fail (we can't get a new blinded UTXO because there are no more unspent asset outputs).

### DIFF
--- a/tests/asset.rs
+++ b/tests/asset.rs
@@ -347,6 +347,7 @@ async fn allow_transfer_one_asset_to_many_beneficiaries() -> Result<()> {
     Ok(())
 }
 
+#[allow(unreachable_code)] // TODO: Remove
 #[tokio::test]
 async fn allow_transfer_assets_to_one_beneficiary() -> Result<()> {
     init_logging();
@@ -539,7 +540,9 @@ async fn allow_transfer_assets_to_one_beneficiary() -> Result<()> {
 
                 info!("Get a second Blinded UTXO");
                 debug!("First was: {blinded_utxo1:?}");
-                let blinded_utxo2 = get_blinded_utxo(&assets_vault_details.assets_change_output)?;
+                let blinded_utxo2 = get_blinded_utxo(&assets_vault_details.assets_change_output);
+                assert!(blinded_utxo2.is_err(), "TODO: Expected error in getting a second blinded UTXO because no new change output has been made");
+                return Ok(());
                 debug!("Blinded UTXO 2: {blinded_utxo2:?}");
 
                 let transfers = vec![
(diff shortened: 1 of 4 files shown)


josediegorobles (approved):
LGTM

tests/asset.rs
```diff
@@ -521,6 +527,72 @@ async fn allow_transfer_assets_to_one_beneficiary() -> Result<()> {
...
+                )
+                .await?;
+
+                debug!("New assets vault details: {assets_vault_details:?}");
+
+                info!("Get a second Blinded UTXO");
+                debug!("First was: {blinded_utxo1:?}");
+                let blinded_utxo2 = get_blinded_utxo(&assets_vault_details.assets_change_output);
+                assert!(blinded_utxo2.is_err(), "TODO: Expected error in getting a second blinded UTXO because no new change output has been made");
+                return Ok(());
```
oacdutra, line 545: This is correct?

tests/asset.rs
```diff
@@ -521,6 +527,72 @@ async fn allow_transfer_assets_to_one_beneficiary() -> Result<()> {
...
+                let transfers = vec![
+                    AssetTransfer {
+                        asset_contract: issued_asset.genesis.to_string(),
+                        asset_utxo: AssetUtxo {
+                            outpoint: assets_vault_details.assets_output.clone().unwrap(),
+                            terminal_derivation: "/0/0".to_string(), // TODO: Should we use /0/1 instead? If so, how? Or should we try to keep it simple for now and reuse addresses?
+                            commitment: "".to_string(),
+                        },
+                        asset_amount: SUPPLY,
+                        change_utxo: assets_vault_details.assets_change_output.clone().unwrap(),
```
oacdutra, line 557: Here breaks because `asset_change_output` is None

oacdutra (commented):
I'm not sure is the return Ok(()) is intencional

tests/asset.rs
cryptoquick: Yes, it was intentional, see the assertion on L544... Also be sure to understand the intention of this PR, as stated in the description.

tests/asset.rs
cryptoquick: I'm not sure a check is needed for this in the test... For multiple transfers to work, new / replacement UTXOs need to be defined upon subsequent calls to this method.

oacdutra (approved):

cryptoquick:
Thanks for the review, @crisdut ! I hope this will help us further down the line.

PR #146, opened 2023-01-24
Add web-accessible sign_psbt method

cryptoquick:


josediegorobles:
Hi the code is good (i have to test it anyway) but I think that is not necessary to have a wasm method appart from that. With this, we are gonna manage this from the extension: we call wasm transfer, then wasm sign and then wasm declare (we don't have wasm declare now).
I think is redundant (and require to load the wallet two times), we can do all from transfer: between the code where we receive the transfer_response and the code where we send the declarerequest code we can sign the psbt. What do you think?

cryptoquick:
Oh, sure, I guess that's another way to do it. I'll make the changes.

cryptoquick:
Oh, no, there's no improvement from doing it together, I just looked. See:
https://github.com/diba-io/bitmask-core/blob/d833cc961c8bce26af79614f21b2f21a4ab6fb1a/src/lib.rs#L560

We're not loading the wallet on the frontend. Besides, this needs to use an xprv, in order to sign it.

josediegorobles:
Ok, it's true, i was thinking with the old ways in mind. So yeah, I think is better like that.

josediegorobles:
Ok, then this code from transfer_assets:

let _declare = async {
        let declare_endpoint = &get_endpoint("declare").await;
        let (_, status) = post_json(declare_endpoint, &transfer_res.declare).await?;
        if status != 200 {
            return Err(anyhow!("Error calling {declare_endpoint}"));
        }
        Ok(status)
    }
    .await?;

Must be outside because has to be declared AFTER the sign. Must be in a declare wasm method and then the extension must call this (using declarerequest from the transfersserializedresponse) after sign the psbt.

Could you do it or better I do it?

josediegorobles (approved):
LGTM

PR #149, opened 2023-01-25
wasm: Add missing ln_check_payment call

grunch:


cryptoquick (approved):
Ah yes, definitely don't forget this!

josediegorobles (approved):

PR #150, opened 2023-01-30
Adding blind to AcceptRequest

josediegorobles:


cryptoquick (approved):
Code looks great. Tested in the extension against the lambdas.

PR #160, opened 2023-03-28
Bump to v0.10

oacdutra:
This PR is an epic to update RGB to v0.10.

**Checklist**
- [x] Finish updating the dependencies
- [x] Add default schema interface (contract base)
- [x] Add RGB contract builder to RGB20 interface
- [x] Create a new endpoint to issue a new RGB20 contract
- [x] Create a new endpoint to generate an RGB invoice
- [x] Allow import RGB20 fungible contract
- [x] Update the PSBT creation operation
- [x] Create a new endpoint to generate a new consignment
- [x] Create a new endpoint to reveal blinded allocations
- [x] Create an official rgb21 interface (RGB-WG/rgb-wallet repo) *** 
- [x] Write a little RGB glossary explain the new terminologies


Cargo.toml
```diff
@@ -45,58 +49,61 @@ payjoin = { version = "0.6.0", features = ["sender"] }
...
+gloo-net = "0.2.6"
+js-sys = "0.3.61"
+serde-wasm-bindgen = "0.5.0"
+wasm-bindgen = { version = "0.2.84", features = ["serde-serialize"] }
+wasm-bindgen-futures = "0.4.34"
+web-sys = "0.3.61"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-amplify = "3.13.0"
+amplify_legacy = { package = "amplify", version = "3.13.0" }
```
cryptoquick: Is there any reason to do legacy here? We should have the same functionality either way, no?

Cargo.toml
```diff
@@ -45,58 +49,61 @@ payjoin = { version = "0.6.0", features = ["sender"] }
 pretty_env_logger = "0.4.0"
 regex = "1.7.0"
 reqwest = { version = "0.11.13", features = ["json"] }
+rgb-std = { version = "0.10.0-rc.1" }
+rgb-wallet = {version = "0.10.0-beta.1" }
```
cryptoquick: Looks like a few of these deps need to be updated already...
rgb-std is in rc.2 now, and rgb-wallet is in rc.1.

src/bin/bitmaskd.rs
```diff
@@ -67,6 +85,7 @@ async fn main() -> Result<()> {
     pretty_env_logger::init();
 
     let app = Router::new()
+        .route("/issue-contract", post(issue_contract))
```
cryptoquick: Please don't make this a new route, just replace the /issue route. It'd reduce maintenance complexity.

src/data/structs.rs
```diff
@@ -84,6 +84,24 @@ pub struct IssueRequest {
...
+    pub description: String,
+    /// Precision of the asset
+    pub precision: u8,
+    /// Amount of the asset
+    pub supply: u64,
+    /// Seal of the initial owner
+    pub seal: String,
+    /// The name of the iface (ex: RGB20)
+    pub iface: String,
+}
```
cryptoquick, line 67: Please just modify the existing struct for issue_asset.

cryptoquick (changes requested):
Just some course corrections if you don't mind.

Cargo.toml
oacdutra: Yes, amplify v4 was released!

The major milestone is a full compatibility with wasm. 

My goal is remove all *_legacy crates after finish the work!

Cargo.toml
oacdutra: Yes, Maxim launch final release today!

I will update ASAP.

src/bin/bitmaskd.rs
oacdutra: I was thinking during the migration, creating routes with a "v2" prefix, so we will be able to remove the previous ones as it is finished. After completing everything, we remove that "v2" prefix.

What do you think?

cryptoquick: No. We haven't even hit 1.0.

oacdutra:
Hi @cryptoquick and @josediegorobles 

I implemented the main functions of RGB. Right now, I'm finishing validating the transfer. After that, I will be working on improvements and testability.

I'm pushing the code for you to follow the work.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ edition = "2021"
 license = "MIT"
 homepage = "https://bitmask.app"
 repository = "https://github.com/diba-io/bitmask-core"
+rust-version = "1.60.0"
```
cryptoquick: Why did you specify the rust-version, btw?

src/lib.rs
```diff
@@ -1,22 +1,66 @@
...
+    psbt::create_psbt as create_rgb_psbt, schemas::default_fungible_iimpl,
+};
+use psbt::ProprietaryKeyDescriptor;
+use psbt::ProprietaryKeyLocation;
+use psbt::ProprietaryKeyType;
+
+use psbt::Psbt;
+use rgbstd::containers::BindleContent;
+use rgbstd::contract::ContractId;
+use rgbstd::contract::GraphSeal;
```
cryptoquick: All this stuff is antipattern... I think it'd be best to move the things using these things and put them in their own modules. They shouldn't be in the top-level lib. In the very least, it should be scoped to just the rgb ops module.

src/lib.rs
```diff
@@ -196,73 +239,168 @@ pub fn list_assets(contract: &str) -> Result<Vec<AssetResponse>> {
...
     let schema_id = contract.schema_id().to_string();
+    let genesis = contract.bindle().to_string();
 
-    Ok(CreateAssetResult {
+    Ok(IssueResult {
         genesis,
-        id,
-        asset_id,
+        id: id.clone(),
+        asset_id: id,
```
cryptoquick: Why are these both the same?

tests/rgb_test_utis.rs
```diff
@@ -0,0 +1,139 @@
+use std::{convert::Infallible, str::FromStr};
```
cryptoquick: This file name is mispelled.

cryptoquick (changes requested):
I appreciate the terminology improvements.
I'll make a PR into this one for adding Carbonado persistence methods.
Aside from that, just a few more questions and comments.

josediegorobles (changes requested):
I'm seeing the code, good work. For now i can't add to the suggestions of @cryptoquick. One thing, there are lots of changes in the API, and that's necessary because the changes in rgb but that means that we will need a little documentation about what needs to be send and what is received from the methods in lib and web. As we are using structs for that, a little description of every field and an example ther could be very useful when the job must be done in the extension (and for other people that will be using bitmask-core)

oacdutra:

I finish the migration to the new version. 

**Some Notes**
- For all interactions between RGB and Blockchain, I implemented with esplora instead of electrum, because esplora had support for wasm32. 
- I needed to add the v0.30 version of rust-bitcoin because of the RGB. I tried to update the whole project for it, but it was getting too much work. So I'll leave that for another time.
- I updated **descriptor-wallet**, **bitcoin_blockchain**, **bitcoin_scripts** and **psbt** crates to using _amplify v4_. This update allows us to use BP-WG libs without custom implementations. After Maxim approves the branches, I will update them to official releases.

cryptoquick (approved):
Really great work so far! This deprecates so much code! I love it.

Next, I'll work on Carbonado support, and then you can use that to build the WASM asset transfer end-to-end test.

cryptoquick:
@josediegorobles I'm going to merge this and additional documentation can be added in later PRs.

josediegorobles:
> @josediegorobles I'm going to merge this and additional documentation can be added in later PRs.

It's ok but next time it's better to wait to have been reviewed by two a so big and with so much changes PR.

PR #162, opened 2023-04-16
Update WASM tests

cryptoquick:
- Update WASM tests.
- Enable WASM tests in CI.
- Refactor to return TransactionDetails instead of just Transaction


oacdutra (approved):
We can merge to development, but I will try resolve conflicts in https://github.com/diba-io/bitmask-core/pull/160 PR, because all related to RGB changed.

cryptoquick:
Sounds good, thanks!

PR #164, opened 2023-04-20
Allow spend inputs with tap(tweak)

oacdutra:
**Description**

Every TapRet-based output of a  RGB transaction has a tweak added to the taproot tree. Unfortunately, no official tooling supports spending this output later today.

This PR allow spend this output type, adding the user field to add the TapRet commitment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -411,7 +411,7 @@ dependencies = [
 [[package]]
 name = "bitcoin_hd"
 version = "0.10.0-alpha.2"
-source = "git+https://github.com/crisdut/descriptor-wallet?branch=release/0.10.0-alpha.2#f021bb329f193c680e29ae9d9c9454938cecb157"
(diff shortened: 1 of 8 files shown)


cryptoquick (approved):
Very nice! Does this put us on the path towards consecutive transfers, then? Ideally we'd have a test for that specifically once that's ready.

Just something to keep in mind in case it's relevant here. Otherwise, it looks great.

oacdutra:
> Very nice! Does this put us on the path towards consecutive transfers, then? Ideally we'd have a test for that specifically once that's ready.
> 

Yes, it's related. Allows us to spend the received bitcoins and assets, but we need fixed the RGB consecutive transfer bug yet.

josediegorobles (approved):
ACK.
It's really good your work @crisdut 
As soon as the RGB bug is corrected we have to have tests for consecutive transfers.

PR #166, opened 2023-04-25
Update version to 0.6.0 beta

josediegorobles:
I think we can publish a beta version yet


cryptoquick (approved):
Very soon. I've got a few final things I need to do tomorrow, but after that, we should be ready to start integrating the new changes very soon.

PR #168, opened 2023-04-27
Carbonado Part 2: Store Harder

cryptoquick:
- [x] Add the capability to specify a file name for what you want the Carbonado file to be called by.
- [x] Replace stock TODOs with calls to Carbonado.

### DIFF
--- a/src/bin/bitmaskd.rs
+++ b/src/bin/bitmaskd.rs
@@ -1,3 +1,4 @@
+#![allow(unused_imports)]
 #![cfg(feature = "server")]
 #![cfg(not(target_arch = "wasm32"))]
 use std::{env, net::SocketAddr};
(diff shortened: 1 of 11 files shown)


src/bin/bitmaskd.rs
```diff
@@ -139,16 +147,17 @@ async fn main() -> Result<()> {
     pretty_env_logger::init();
 
     let app = Router::new()
+        /* TODO: ECDH bearer auth.
```
oacdutra, line 150: Doubt, how will this work?

oacdutra (approved):

src/bin/bitmaskd.rs
cryptoquick: What, ECDH bearer auth? Yes, we already do something similar with carbonado_node.

If you're asking if it's okay to comment out the node methods, yes, this is fine for now, since with this, RGB and Carbonado all work pretty much entirely clientside, with zero server calls except for Carbonado persistence and Electrum.

cryptoquick: Oh, I misread your question. It's been a long day. I'll show you how tomorrow.

josediegorobles (approved):
ACK

PR #169, opened 2023-04-27
Transfer Assets - Part 1 (Support to RgbWallet)

oacdutra:
**Decription**

This PR intent add initial support to RgbWallet, the structure to manage UTXOs and Tapret Commitments. This is a alternative to Amount Protocol.

Also, in the part 2, I will start implement all Asset Transfers tests.

**Tasks**
- [x] Support to RgbWallet (alternative to amount protocol).
   - [x] Retrieve commitments from PSBT
   - [x] Save commitments
   - [x] List UTXOs per contract
   - [x] List addresses used per contract
   - [x] List addresses used per contract
   - [x] Allow get next unused Address
   - [x] Allow get next unused UTXO
   - [x] Add esplora support
- [x] Integration Tests
  - [x] Setup docker to each integration tests
  - [x] Execute "send coins" directly to the node

**Other Tasks**
 - [x] Add import contract method


src/bin/bitmaskd.rs
```diff
@@ -203,6 +217,7 @@ async fn main() -> Result<()> {
         .route("/contracts", get(contracts))
         .route("/interfaces", get(interfaces))
         .route("/schemas", get(schemas))
+        .route("/imports", post(imports))
```
cryptoquick: Can we please change this to /import, singular, not plural?

src/rgb/imports.rs
```diff
@@ -0,0 +1,28 @@
+use amplify::{confinement::Confined, hex::FromHex};
```
cryptoquick: Change this filename from plural to singular ("import.rs")

src/rgb/resolvers.rs
```diff
@@ -1,9 +1,19 @@
...
+use bitcoin::Script;
+use bitcoin_hashes::hex::FromHex;
 use bp::{LockTime, Outpoint, Tx, TxIn, TxOut, Txid, VarIntArray};
 use futures::executor;
+use rgb::{
+    prelude::{DeriveInfo, MiningStatus},
+    Utxo,
+};
 use rgbstd::resolvers::ResolveHeight;
 use rgbstd::validation::ResolveTx as ResolveCommiment;
```
cryptoquick:
```suggestion
use rgbstd::{resolvers::ResolveHeight, validation::ResolveTx as ResolveCommiment};
```

src/rgb/structs.rs
```diff
@@ -1,5 +1,7 @@
 use bitcoin::Address;
-use std::str::FromStr;
+use bitcoin_scripts::address::AddressCompat;
+use rgb::{persistence::Stock, RgbWallet, TerminalPath};
+use std::{collections::HashMap, str::FromStr};
```
cryptoquick: Put stdlib imports at the top of the file, separate from the crates imports

src/structs.rs
```diff
@@ -46,6 +46,7 @@ pub struct PublicWalletData {
     pub nostr_npub: String,
     pub xprvkh: String,
     pub xpubkh: String,
+    pub xpub: String,
```
cryptoquick, line 49: What is this xpub?

cryptoquick (changes requested):
Very good work so far. Added a few review comments.

One question;

Can't we just use the same descriptors we use with BDK with RGB wallet also?

It's fine if we don't have an RGB integration test, the utils you build are really clever, but I still want those. They can be done in a separate PR, though.

src/structs.rs
oacdutra: I will need plain xpub to build RgbWallet.

cryptoquick: Interesting... Good to know.

cryptoquick (approved):
All issues addressed. ACK.

cryptoquick:
@josediegorobles Armando needs this merged, but let him know if you see any issues

PR #173, opened 2023-05-04
Transfer Assets - Part 2 (Make first e2e transfer operation)

oacdutra:
This PR intends to add all integration tests to support transfer assets operation.

**Tasks** 
- [x] Add RGB Watcher
- [x] Add Integration Tests (Carbonado and RGB v0.10)
  - [x] Dockenized `bitmaskd`
  - [x] Test Import Contracts
  - [x] Test Create invoice
  - [x] Test Pay Invoice
  - [x] Test Accept Transfer
- [x] Add WASM Compatibility
  - [x] Add RGB Resolvers prefetch strategy
  - [x] Test import contract (example)

**Others**
- [x] Bump dependencies (with RGB v10.1)

*** Some transfers operation does not work yet. See below:
* Issuer sends consecutive transfers to beneficiaries;
* Beneficiary (New Owner) sends consecutive transfers to beneficiaries;
* All contracts anchored in the same UTXO are updated after the `accept` operation ;


oacdutra:

At first, I completed the implementation, but I'm having problems in one case:

For some reason, I get [this bug](https://github.com/diba-io/bitmask-core/actions/runs/4913365897/jobs/8773476395?pr=173#step:8:108) when I compile Esplora in the blocking feature enabled with the wasm32 target. But I need to use sync-esplora to run the RGB resolvers.

I tried running the async-esplora version using futures-rs, and the compilation works. But none of the calls to esplora worked.


Could you guys help me to run Esplora async "synchronously"?

To run the tests locally, just start docker and run the commands below:

```bash
docker-compose up -d node1
docker-compose up -d node1 bitmaskd
docker-compose exec -T node1 cli loadwallet default
docker-compose exec -T node1 cli -generate 500
```


Thanks

src/rgb/resolvers.rs
```diff
@@ -27,14 +25,20 @@ impl rgb::Resolver for ExplorerResolver {
         scripts: BTreeMap<DeriveInfo, bitcoin_30::ScriptBuf>,
     ) -> Result<BTreeSet<rgb::prelude::Utxo>, String> {
         let mut utxos = bset![];
-        let explorer_client = EsploraBlockchain::new(&self.explorer_url, 100);
+        // TODO: Find a way to run async function synchronously (wasm32)
```
oacdutra: @cryptoquick and @josediegorobles, 

here is the place I use the esplora-sync.

josediegorobles: That's really strange, it supposed to be running the blocking version yet. I will try if i can figure out myself something.

cryptoquick:
This blocking feature uses ureq, which does not appear to have any kind of wasm support, and that makes sense, since it makes TCP and UDP connections, with no conditional compilation directives utilized for the wasm32 target.

It's okay, it looks like this can be solved. After diving in, the object-oriented model you're using is not serving us well. I'll make a PR into this one that simplifies things.

oacdutra:
> This blocking feature uses ureq, which does not appear to have any kind of wasm support, and that makes sense, since it makes TCP and UDP connections, with no conditional compilation directives utilized for the wasm32 target.
> 
> It's okay, it looks like this can be solved. After diving in, the object-oriented model you're using is not serving us well. I'll make a PR into this one that simplifies things.

I solved the problem of resolvers. I added a prefetch layer that feeds the resolvers. This ensures that we don't need to make any modifications.

Tomorrow I will finalize the tests and add the pending fields in the answers.

src/rgb/psbt.rs
```diff
@@ -66,6 +66,7 @@ pub fn create_psbt(
         .map(|AddressAmount { address, amount }| (address.script_pubkey().into(), amount))
         .collect();
 
+    let descriptor_pub = descriptor_pub.replace("/20/*", "/*/*");
```
cryptoquick: We can probably just change the constant, no?

src/rgb/resolvers.rs
```diff
@@ -1,40 +1,48 @@
...
 
+#[derive(Default)]
 pub struct ExplorerResolver {
     pub explorer_url: String,
+    // Prefetch Data (wasm32)
+    pub utxos: BTreeSet<Utxo>,
+    pub next_utxo: String,
+    pub txs: HashMap<bitcoin::Txid, bitcoin::Transaction>,
+    pub bp_txs: HashMap<Txid, Tx>,
 }
```
cryptoquick, line 23: Very interesting... Very clever solution. I worry this might return false negatives if the transfer just took place, but we can work around that, yes?

cryptoquick (changes requested):
Looking very good so far! Just a couple of questions.

src/rgb/psbt.rs
oacdutra: Yes, I'm going to start the review, and I'm going to replace that check with one of the method's parameters.

src/rgb/resolvers.rs
oacdutra: Yes, we can.

For now, I only add transactions to the hashmap if they actually exist in the blockchain/mempool, so I think it should work.

src/rgb/psbt.rs
cryptoquick: Sounds good!

src/rgb/resolvers.rs
cryptoquick: That could work!

oacdutra:
@cryptoquick,

Could you put the address of the `$RGB_VAULT_ADDRESS` environment variable, please?


I added the instructions in the README.

oacdutra:
There is a test that performs all the steps, it's [here](https://github.com/diba-io/bitmask-core/pull/173/files#diff-24bc0743c8c91129b89f66e2950f330fdee603d1585f0b3b2ad35662af277eeeR80-R105)


Also, @cryptoquick can you help me fix this test?

https://github.com/diba-io/bitmask-core/actions/runs/4933240675/jobs/8817024267?pr=173#step:7:176

src/bitcoin/keys.rs
```diff
@@ -49,6 +50,18 @@ fn xpub_desc(xprv: &ExtendedPrivKey, path: &str, change: u32) -> Result<String>
...
+    let secp = Secp256k1::new();
+    let xprv = get_descriptor(xprv, path, change)?;
+    let xpub = xprv.to_public(&secp)?;
+
+    if let DescriptorPublicKey::XPub(desc) = xpub {
+        Ok(desc.xkey.to_string())
+    } else {
+        Err(anyhow!("Unexpected xpub descriptor"))
+    }
+}
```
cryptoquick: Oh wow, so, it looks like this generates a different xpub than we were using before.

We're actually really lucky my test caught this problem, this could have made it so everyone's wallet addresses were different than they were before.

Can you explain why this change is needed?

cryptoquick (changes requested):
I've diagnosed the problem, it was easy to correct, I'll open a PR in to this one to resolve it.

src/bitcoin/keys.rs
oacdutra: Sure.

I created this method because I needed the public key to be the same as the key represented in the descriptors.

Today every public key in descriptor is based on BTC_PATH (e.g. m/86h/1h/0h) (see [here](https://github.com/diba-io/bitmask-core/blob/4558665c588c9dee5402d7913610a47e8d1d4a18/src/bitcoin/keys.rs#L22-L36)).

However, the initial xpub is generated without this path (see [here](https://github.com/diba-io/bitmask-core/blob/4558665c588c9dee5402d7913610a47e8d1d4a18/src/bitcoin/keys.rs#L107-L111)).

In case you need this key without BTC_PATH, I can create another field called "watcher_xpub", looking something like this:

```rust
    let public = PublicWalletData {
        btc_descriptor_xpub,
        btc_change_descriptor_xpub,
        rgb_assets_descriptor_xpub,
        rgb_udas_descriptor_xpub,
        nostr_pub,
        nostr_npub,
        xprvkh,
        xpubkh,
        xpub: xpub.to_string(),
        watcher_xpub: xpub_plain,
    };
```

cryptoquick: So, is the reason you can't just use the btc_descriptor_xpub simply because it has `tr(...)` string formatting?

oacdutra: Yes, wacher only needs xpub as it uses a tapret based descriptor. In order not to run a regex to extract xpub every time, I create this field.


If you prefer, I create a new field called watch_xpub, as I will be using it for all contracts in the future.

cryptoquick: That works! I'll close my PR.

src/structs.rs
```diff
@@ -139,8 +165,26 @@ pub struct ImportRequest {
 pub struct ImportResponse {
     /// The contract id
     pub contract_id: String,
-    /// The contract interfaces
-    pub ifaces: Vec<String>,
+    /// The contract impl id
+    pub iimpl_id: String,
```
cryptoquick, line 169: Are these fields misspelled?

README.md
```diff
@@ -56,17 +56,18 @@ For running bitmask-core tests in Regtest Mode, please follow the steps below:
...
-7. Run the test to get main address: `cargo test --test wallet -- create_wallet --exact`.
+6. Create an environment variable called **TEST_WALLET_SEED** with mnemonic generated in the **step 5** (only wasm32).
+7. Run the test to get main address for bitcoin and rgb: `cargo test --test wallet -- create_wallet --exact`.
 8. Load your wallet in the bitcoin node: `node1 loadwallet default`.
 9. Generate new first 500 blocks: `node1 -generate 500`.
-10. Send some coins to the main wallet address: `node1 sendtoaddress {ADDRESS} 10`. Change `{ADDRESS}` with the address generated in the **step 7**.
-11. Mine a new block: `node1 -generate 1`
-12. Run the test to check the balance: `cargo test --test wallet -- get_wallet_balance --exact`.
+10. Send some coins to the main wallet address: `node1 sendtoaddress {BTC_ADDRESS} 10`. Change `{BTC_ADDRESS}` with the address generated in the **step 7**.
+11. Send some coins to the rgb wallet address: `node1 sendtoaddress {RGB_ADDRESS} 10`. Change `{RGB_ADDRESS}` with the address generated in the **step 7**.
```
cryptoquick: Are these not the same environment variables as those used here?
https://github.com/diba-io/bitmask-core/pull/173/files#diff-9ed853ee5016b3f7cb01ee3da6506fd9e926a923324f327507c34e64f83dfe10R131-R132
If so, be sure to make sure they match.

src/bitcoin.rs
```diff
@@ -17,6 +19,7 @@ mod psbt;
 mod wallet;
 
 use crate::structs::EncryptedWalletDataV04;
+use crate::structs::{SignPsbtRequest, SignPsbtResponse};
```
cryptoquick:
```suggestion
use crate::structs::{EncryptedWalletDataV04, SignPsbtRequest, SignPsbtResponse};
```

src/bitcoin/psbt.rs
```diff
@@ -29,8 +29,9 @@ pub async fn sign_psbt(
         let txid = tx.txid();
         let tx = blockchain
             .get_tx(&txid)
-            .await?
-            .expect("tx that was just broadcasted now exists");
+            .await
+            .expect("tx that was just broadcasted now exists")
+            .unwrap();
```
cryptoquick: Can this `unwrap` not be replaced with a `?` to bubble up errors?

src/rgb/carbonado.rs
```diff
@@ -1,10 +1,13 @@
 use amplify::confinement::{Confined, U32};
 use anyhow::Result;
 use rgbstd::persistence::Stock;
+use serde_json::to_vec;
 use strict_encoding::{StrictDeserialize, StrictSerialize};
 
 use crate::carbonado::{retrieve, store};
 
+use super::structs::RgbAccount;
```
cryptoquick: Let's pull this from `crate::`
Also, please check other places the super pattern has been used and see if we can't reduce the different patterns that are needed for consistency's sake.

src/rgb/prefetch.rs
```diff
@@ -0,0 +1,239 @@
...
+    wallet: &mut RgbWallet,
+) {
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub async fn prefetch_resolve_txs(txids: Vec<Txid>, explorer: &mut ExplorerResolver) {}
+
+#[cfg(target_arch = "wasm32")]
+pub async fn prefetch_resolve_commit_utxo(contract: &str, explorer: &mut ExplorerResolver) {
+    let esplora_client: EsploraBlockchain = EsploraBlockchain::new(&explorer.explorer_url, 100);
```
cryptoquick: Do you think we could use this method, and set the concurrency value to six? IIRC, browser pages can make up to six requests in parallel.
https://docs.rs/bdk/latest/bdk/blockchain/esplora/struct.EsploraBlockchain.html#method.with_concurrency

Also, let's move this to a function so any changes can be made in one place.

src/rgb/wallet.rs
```diff
@@ -1,3 +1,4 @@
+use crate::structs::{AllocationDetail, WatcherDetail};
```
cryptoquick: This is misplaced, be sure to put this below the extern crate imports.

cryptoquick (changes requested):
This is fantastic! I just went over it through a fine-toothed comb, and I have some small stylistic nits.
Also, a couple things to be kept in mind (though let's address them later):

We really need to transition the project to using a stronger error type system, ideally using something like thiserror. That might be low enough priority to do after the BTC23 launch, however.
Just be aware, everywhere unwrap and expect are used can cause a runtime panic.

I'm glad you've tested many more things here. We'll definitely need another follow-up PR testing more extensive workflows, however.

src/structs.rs
oacdutra: No, this acronym is to interface implementation.

josediegorobles (approved):
LGTM

src/structs.rs
cryptoquick: Good to know, thank you!

cryptoquick (approved):
Awesome! The new changes look good to me!

PR #177, opened 2023-05-09
v4 descriptor migration

cryptoquick:
I tried adding v3 migration, but it encoded the descriptor like this: `[86,129,171,148,249,162,55,222,139,140,77,156,203,5,172,47,246,135,31,218,172,220,168...]`

I figured that was sufficiently ancient that very few people are probably running that or upgrading. I could add that, but only if it's requested by one of our users.


josediegorobles (approved):
LGTM

PR #178, opened 2023-05-09
Add v5 migration test

cryptoquick:
I needed to test v5 migrations to see if that was an issue, but it seems like the same code works fine. I still refactored it in case more migrations need to be added in the future.

Also, I did investigate v3 migrations, but it turns out, we did not encrypt the mnemonic into the descriptor at the time:
https://github.com/diba-io/bitmask-core/blob/72aa74c8fa8132629d5adc53640f177cbdf679ff/src/lib.rs#L31-L39

In addition, those were SegWit-only wallets, so none of the funds would be accessible anyway. Users can migrate their funds to a new wallet using https://segwit.bitmask.app


josediegorobles (approved):
LGTM

cryptoquick:
ty! :pray:

PR #184, opened 2023-05-12
Move xprvkh

cryptoquick:
So, I forgot we were using xprvkh as a password for LNDHubX, and so I had miscategorized it when I had split up the two.


josediegorobles (approved):
ACK

PR #187, opened 2023-05-13
Fee and TX Improviments

oacdutra:
This PR intent adds some improvements based on @josediegorobles feedback. See below:

#### **Genesis Data**
Now, the methods `create_contract`, `import_contract`, and `list_contracts` return the new field called `GenesisFormats`. This field has genesis in legacy and strict formats, like `ContractFormats`.

**Important:** Even with a similar format, contracts and genesis from version 0.10 are incompatible with previous versions.

#### **Fee Calculator**

Currently, We use the create psbt method by descriptor-wallet, which does not support fee calculations. 

To solve that, I using **bdk first** to get the fee, and then I use the fee result in the wallet descriptor psbt generation. This approach has solved the fee problems by now, but I would like to return to this problem and discuss it with you guys.

**PS:** This was the solution to the extension problem not calculating the fee.

#### **SignPsbtResponse**

Now, the **psbt sign** response returns the transaction (in hexadecimal format).

### DIFF
--- a/src/bin/bitmaskd.rs
+++ b/src/bin/bitmaskd.rs
@@ -55,7 +55,7 @@ async fn invoice(
     Ok((StatusCode::OK, Json(invoice_res)))
 }
 
-async fn psbt(
+async fn _psbt(
     TypedHeader(auth): TypedHeader<Authorization<Bearer>>,
     Json(psbt_req): Json<PsbtRequest>,
 ) -> Result<impl IntoResponse, AppError> {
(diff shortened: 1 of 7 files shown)


cryptoquick (changes requested):
Very neat! Be sure to add the fee estimation methods to the RGB module of web.rs so we can make use of them on the JS side.

josediegorobles:
LGTM. With the Fe estimation methods that @cryptoquick says, I would be happy with this PR

cryptoquick (approved):
After discussing this with Armando, he'll do the fee estimation as separate work.

Issue #189, opened 2023-05-13
Hex and base64 codec methods


cryptoquick:
In carbonado.rs:
- [ ] `encode_base64(bytes: &[u8]) -> String`
- [ ] `encode_hex(bytes: &[u8]) -> String`
- [ ] `decode_base64(string: &str) -> Vec<u8>`
- [ ] `decode_hex(string: &str) -> Vec<u8>`

Don't forget to export these methods in web::carbonado.

oacdutra:
@cryptoquick, for some reason, not close this issue after your merge. Can I close?

PR #192, opened 2023-05-14
Separate password hashing from encryption

### LINKED ISSUE #190: Separate password hashing from encryption
In the extension, when we switch networks, we would rather hash the password once and keep the hash in memory, than ask the user for their password every the time it's needed. To do this, however, we need to separate password hashing from encryption and decryption, so the password can be hashed just once, and then the hash is provided to the descriptor decryption methods.

cryptoquick:
Closes #190 #193 #189

I realize this will essentially make all previous descriptor formats outdated, so I think we should have a site where we host an old version, BitMask 0.5.2, and people can migrate their testnet sats from there.

I've also added Argon2id hashing and Carbonado storage encoding helper methods.


josediegorobles (approved):
LGTM

PR #194, opened 2023-05-14
Fix misnamed method and args in web.rs

cryptoquick:
- Fix misnamed method in web.rs, get_mnemonic_seed to new_mnemonic_seed, to match the bitcoin module method it's calling.
- Fix misnamed arguments in web.rs to save_mnemonic_seed.
- Bump version to 0.6.0-beta.7

I discovered this when integrating the new extension API into bitmask-wallet.


josediegorobles (approved):
LGTM

PR #195, opened 2023-05-16
FIX: Expose Contract Data

oacdutra:
This PR intent show contract data. 

Problem reported by @josediegorobles.

PS: To obtain the balance, it is necessary to create the watcher for the user, using the watcher_pub field.

### DIFF
--- a/src/rgb/contract.rs
+++ b/src/rgb/contract.rs
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-use amplify::hex::ToHex;
+use amplify::{confinement::Confined, hex::ToHex};
 use bech32::{encode, ToBase32};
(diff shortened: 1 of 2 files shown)


cryptoquick (approved):
Interesting. This solution for reading contract fields is a little unwieldy. I'm wondering what @dr_orlovsky thinks. This works for now, though.

oacdutra:
> Interesting. This solution for reading contract fields is a little unwieldy. I'm wondering what @dr_orlovsky thinks. This works for now, though.

I thought the same when I saw it for the first time, the new version is easier to navigate through the data, I'm testing it now.

josediegorobles (approved):
LGTM

PR #196, opened 2023-05-17
Password hash and encryption test

cryptoquick:
This is just to sanity-check an issue found in the extension.


josediegorobles (changes requested):
It's ok but you have to solve the conflicts

cryptoquick:
Good catch! Done!

josediegorobles (approved):
ACK

Issue #181, opened 2023-05-10
Additional RGB transfer integration tests


cryptoquick:
We need more tests for more extensive workflows.

It will be helpful to have a graphic outlining the various token transfer flows.

Issue #140, opened 2023-01-18
Consecutive transfers


Gigi3d:
ACs:

- [ ] Can transfer a token multiple times, one after another (consecutively)
- [ ] Can create a new receive outpoint after the spend outpoint is spent
- [ ] Can track multiple asset outputs in the wallet so it can be recovered from a mnemonic

cryptoquick:
I'm not sure we have well-defined behavior around this problem. Conceptually, the vault is just a set of UTXOs we can mint and receive assets on, and so funding the vault might not be the right name for this; instead, maybe "create vault" or "start vault" ?

It's not really an account per se, the vault is not an address; it's a UTXO that can, for a time, receive assets, unless it's spent, then it can no longer receive assets, and all assets are transferred to a new one. Assets shouldn't be lost if sent to a spent tx output, since transfer is an atomic operation.

Presently, we don't handle making new change UTXOs well after a transfer (multiple transfers of a token, one after the other), but I think it's kind of assumed that the wallet will use any sats in the main bitcoin on-chain wallet as-needed. There is also a related question of how we order and persist outputs for a new change UTXO.

josediegorobles:
I think Create vault is a better description and yeah I don't see necessary notifications of it running low. The sats necessaries for change must be take from the main wallet.
The point made by @cryptoquick are correct and necessary to have a good management of the vault

Gigi3d:
This is way  better and less complicated than I thought. We can close this issue i presume?

cryptoquick:
I'd prefer to repurpose this issue, since it effectively encapsulates some thinking we still need to do in order to support multiple transfers.

cryptoquick:
I'll be renaming the vault details so that the receive outpoints have been renamed spend, and change outpoints renamed to receive. Then, I think there might be a way to determine which outpoint is which just by looking at past transactions for each asset vault wallet, we'd just need to do the funding transactions in a different way. If we do them with the old output, plus using an output from the bitcoin wallet for fees, then we can create a new receive output, and then an output sending the remainder back as change to the latest bitcoin wallet change address. This might work... I won't know unless I try, though.

We will know which outpoint is which through the convention that the transaction prior's receive outpoint is now the new spend outpoint, and the new receive outpoint is the outpoint that does not belong to the bitcoin wallet, according to this: https://docs.rs/bdk/latest/bdk/wallet/struct.Wallet.html#method.get_tx

[...]

PR #199, opened 2023-05-18
Support UDA part1

### LINKED ISSUE #114: RGB21 support
We'll need to update this in order to support RGB21 for UDAs in addition to RGB20 for fungible assets.

oacdutra:
**Details**
- [x] RGB Contracts v0.10.2
- [x] Issue
- [x] Import
- [x] Invoice
- [x] Transfer
- [x] Accept
- [x] https://github.com/strict-types/strict-types/pull/10
- [x] Issuer Balance (descriptor-wallet bug)
- [x] Support Media (rgb-wallet bug)

Closes #114


oacdutra:

Now we have support to RGB21 (UDA). I need make some changes in rgb-wallet (I will open a PR in rgb-wallet soon).

BTW, we have two issues yet:

* For some reason, the structs **MediaType** and **MediaRegName** does not working when we try validate the global state of the contract.

* I dont kwon why this error occurs: https://github.com/diba-io/bitmask-core/actions/runs/5020876340/jobs/9002747007#step:9:311

josediegorobles (approved):
ACK

cryptoquick (changes requested):
So, the error appears to be coming from the changes in the web asset test, so it might be a legitimate error? Bear in mind, this error happens at runtime, not compile time:
Error: invalid type: map, expected a string at line 1 column 68

It might be helpful to build the same test, web_asset, in x86_64 too, maybe?

oacdutra:
> So, the error appears to be coming from the changes in the web asset test, so it might be a legitimate error? Bear in mind, this error happens at runtime, not compile time: Error: invalid type: map, expected a string at line 1 column 68
> 
> It might be helpful to build the same test, web_asset, in x86_64 too, maybe?

Hi guys,

I found this bug.

For some reason, the latest version of rgb smart contract StandardTypes causes `RuntimeError: memory access out of bounds` when compiled in wasm32 target.

I will need more time to check that.

My first strategy will be to create a precompiled form of the standard, because the new version does strict type transpilation in runtime and I think this causes memory overflow.


Issue related: https://github.com/RGB-WG/rgb-wallet/issues/52

cryptoquick:
That sounds awesome if it's doable.

oacdutra:
> That sounds awesome if it's doable.

bug fixed!

oacdutra:

I found the issuer bug:

The wallet descriptor always forces the "change address" derivation to  /1/* , not what we determine in the input (asset: /20/* or udas: /21/*).

I will resolve that tomorrow because I need to make more tests in descriptor-wallet and rgb-wallet "velocity" verification.

cryptoquick:
Good thinking! I look forward to seeing it.

oacdutra:
Guys,

I finished!!

I needed to do a lot of modifications in the dependencies, and I will still this week, send all the PRs for official releases. This is important to avoid custom contract interface implementations.

One thing that worry me are possible breaks of compatibility in the contract interfaces. I intend to open some discussions on RGB-WG about this topic, to plan some work related that!

BTW,  my next steps: #181 and #140

cryptoquick:
Awesome! I think we should just merge this since it's passing tests, do a new release of bitmask-core, see how well that integrates with the latest, and we're good to go from there. What do you think, @josediegorobles ?

cryptoquick (approved):
ACK

josediegorobles (approved):
ACK

PR #202, opened 2023-05-24
Server for the marketplace

josediegorobles:


.env
```diff
@@ -19,4 +19,5 @@ NODE_HOST=https://qvijq4x0ei.execute-api.us-east-2.amazonaws.com/dev
 # NODE_HOST=http://127.0.0.1:7070
 LNDHUB_ENDPOINT=https://lndhubx.bitmask.app
 CARBONADO_ENDPOINT=http://localhost:7070/carbonado
-BITCOIN_NETWORK=regtest
+BITCOIN_NETWORK=regtes
```
oacdutra:
```suggestion
BITCOIN_NETWORK=regtest
```

src/bin/bitmaskd.rs
```diff
@@ -27,18 +28,57 @@ use bitmask_core::{
...
+    let sk = issuer_keys.private.nostr_prv;
+
+    let issue_res = issue_contract(&sk, issue).await?;
+
+    Ok((StatusCode::OK, Json(issue_res)))
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SelfIssueRequest {
```
oacdutra: Could you move the `SelfIssueRequest` to src/structs.rs, please?

oacdutra (changes requested):
This is good!

Please, check my comment.

src/bin/bitmaskd.rs
```diff
@@ -27,18 +28,57 @@ use bitmask_core::{
...
 
-    let nostr_hex_sk = auth.token();
+    let issuer_keys = save_mnemonic(&get_marketplace_seed().await, "").await?;
+
+    let sk = issuer_keys.private.nostr_prv;
+
+    let issue_res = issue_contract(&sk, issue).await?;
+
+    Ok((StatusCode::OK, Json(issue_res)))
+}
```
cryptoquick, line 46: I'm really worried about the security implications of this... Please secure this with a bearer token, similar to here:

https://github.com/diba-io/bitmask-core/pull/202/files#diff-69dce2abeb3a3dbe31441f8e383f2c9ed664335f5293f0c94fd0ad22ba53c080R255

We can then put the auth token in the server's env vars, and then it can be provided by the server calling it.

cryptoquick (changes requested):
Requesting change regarding security...

src/bin/bitmaskd.rs
josediegorobles: I'm with you but i'm going to wait until all is working because is much faster to test it like that. So when the hut8 server is working properly i will do another commit with this request.

cryptoquick: Sounds good

src/bin/bitmaskd.rs
josediegorobles: Yeah

.env
```diff
@@ -4,19 +4,21 @@
...
 # BITCOIN_ELECTRUM_API_MAINNET=mempool.space:50001
 # BITCOIN_ELECTRUM_API_TESTNET=mempool.space:60001
 # BITCOIN_ELECTRUM_API_SIGNET=mempool.space:60601
-NODE_HOST=https://qvijq4x0ei.execute-api.us-east-2.amazonaws.com/dev
+NODE_HOST=://hut8-ca-on-mississauga.carbonado.io/carbonadoapi.us-east-2.amazonaws.com/dev
 # NODE_HOST=http://127.0.0.1:7070
 LNDHUB_ENDPOINT=https://lndhubx.bitmask.app
 CARBONADO_ENDPOINT=http://localhost:7070/carbonado
-BITCOIN_NETWORK=regtest
+BITCOIN_NETWORK=bitcoin
```
oacdutra: I'm not sure, but in most places Network conversion expects mainnet not bitcoin.

```suggestion
BITCOIN_NETWORK=mainnet
```

josediegorobles: I updated .env by mistake...

PR #207, opened 2023-05-25
Calling itself bypass

josediegorobles:
Trying that the server call itself


cryptoquick:
I'm skeptical this will fix the issue, but it might be useful in ruling some things out.
Some of the .env fixes might be useful, though.

PR #209, opened 2023-05-25
Adding full support to assets storing and udas issuing on the server

josediegorobles:
- Assets storing on the server
- UDAs issuing on the server (for marketplace)


cryptoquick (approved):
I'm really happy with how these changes are shaking out. I'm good to merge if you are.

PR #214, opened 2023-05-28
Fix watcher inconsistencies

oacdutra:
This PR fix some issues and add new feature: stress tests module.

### **Fixies**
1.  On issue operations, the watcher was not saving.
2.  On contracts listing operations, the watcher only loaded data related to Fungibles.

### **Changes**
1.  Now, a watcher is only created when it doesn't exist..

### **Features**

I added a new module for stress tests, where we create over 400 contracts at once and check contract allocations and balances. The test is divided by fungibles and UDAs, where we test:

- Issue of 150 contracts in a single UTXO
- Issuing 26 contracts while the same portfolio is receiving sats, that is, producing new UTXOs.
- Issuing 26 contracts with the same portfolio while we are spending UTXOs, to verify that the portfolio is managing UTXOs correctly.

In the future, I intend to extract this module to test other features of the wallet, such as LN, payjoin, etc... always pointing to regtest.

#### How this Works

We need change the variables `STRESS_TEST` to **true** and `BITCOIN_NETWORK` to `regtest`.

**Important:** Running stress tests ONLY regtest, because we spent a lot of sats!

**Review**

@josediegorobles, I noticed that marketplace contracts are being anchored to a single UTXO. We have to be careful because the current version of RGB has some problems in these situations:

- For now, we are unable to carry out change-based transfers, whether the contract determined in the invoice or the other contracts anchored to the UTXO.

- Stock was not prepared to support the transition change of all contracts linked to a single UTXO, that is, we may lose funds if two or more contracts are anchored to a single UTXO.


crisdut:
Hi @josediegorobles and @cryptoquick 

I need some help here:

After the change in carbonado storage (addition of the "server" flag), all my tests started to break, as they were dependent on bitmaksd in docker.

Now it "locally saves the carbonado" and tries to search inside docker.

Would you like more details on the reason for this change?

josediegorobles (approved):
ACK, we can fix tests in the original PR

PR #215, opened 2023-05-28
Support Collectibles - part1

oacdutra:
### **Description**

This PR intent support to collectibles using RGB21 interface.

#### **BMC Changes**
I needed made some changes to support collectibles, check the list bellow:
- IssueRequest->MediaInfo[] is now IssueRequest->IssueMetaRequest
- IssueMetaRequest support two types of UDA: Single UDA and Collectible
- ContractResponse->MediaInfo[] is ContractResponse ->ContractMeta
- ImportResponse->MediaInfo[] is ImportResponse ->ContractMeta
- ContractMeta support two types of UDA: Single UDA and Collectible
- AllocationDetail->value is now AllocationValue
- AllocationValue support two types of Value: u64 and UDAPosition(TokenIdex, Fraction)

#### **Single UDA vs Collectibles (RGB)** 

The big difference between the two types of UDA is the change in the RGB21 interface. The current version was incompatible with the LNPBP-21 specification because the interface spec supports adding multiple tokens, while the implementation does not.

But I intend to validate with Maxim. So for now, let's keep this PR as a draft until he answers my questions.

### **Highlights**

- [x] Issue Collectible
- [x] Import Collectible
- [x] Invoice Random Collectible Item
- [x] Invoice Specific Collectible Item
- [x] Transfer Collectible
- [x] Accept Collectible

**PS:** Please check the tests/rgb/collectibles.rs to understand the changes.


cryptoquick (approved):
Looked over the code and tests, this looks like really sold work, thank you so much, @crisdut

oacdutra:
> Looked over the code and tests, this looks like really sold work, thank you so much, @crisdut

Thanks @cryptoquick 

I'll keep this PR as a draft until Maxim clears my doubts, ok?

src/structs.rs
```diff
@@ -127,12 +130,45 @@ pub struct SelfIssueRequest {
...
     pub description: String,
-    /// attachments and media (only RGB21/UDA)
-    pub medias: Option<Vec<MediaInfo>>,
+    /// contract metadata (only RGB21/UDA)
+    pub meta: IssueMetaRequest,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct IssueMetaRequest(pub Vec<IssueMetadata>);
```
cryptoquick, line 139: Why does Issue have multiple Metadata but Contract only has one?
https://github.com/diba-io/bitmask-core/pull/215/files#diff-14138424001ab746459e9159716ae9e8acaca97ca7c5ac2020cb2bcf18a47cecR280

cryptoquick (changes requested):
After digging into this more, I think it might be helpful if we didn't separate UDAs and collectibles; I think all UDAs should be capable of belonging to a collection.

josediegorobles:
> After digging into this more, I think it might be helpful if we didn't separate UDAs and collectibles; I think all UDAs should be capable of belonging to a collection.

That's the better. Always we can hava a "no collection" collection

josediegorobles:
now is this: https://github.com/diba-io/bitmask-core/pull/217

PR #224, opened 2023-05-30
Fix web storage

oacdutra:
Alternative to #225 and #223


src/carbonado.rs
```diff
@@ -114,12 +114,20 @@ pub async fn retrieve(sk: &str, name: &str) -> Result<Vec<u8>> {
...
     let pk = public_key.to_hex();
 
-    let filepath = handle_file(&pk, name, 0).await?;
+    let network = NETWORK.read().await.to_string();
+    let mut final_name = name.to_string();
+    if !name.contains(&network) {
+        final_name = format!("{network}-{name}");
+    }
+
+    let filepath = handle_file(&pk, &final_name, 0).await?;
```
oacdutra, line 130: I'm not sure if this is correct!

src/carbonado.rs
```diff
@@ -132,11 +140,19 @@ pub async fn retrieve(sk: &str, name: &str) -> Result<Vec<u8>> {
...
 #[cfg(feature = "server")]
 pub async fn handle_file(pk: &str, name: &str, bytes: usize) -> Result<std::path::PathBuf> {
+    use crate::constants::NETWORK;
+
+    let network = NETWORK.read().await.to_string();
+    let mut final_name = name.to_string();
+    if !name.contains(&network) {
+        final_name = format!("{network}-{name}");
+    }
+
```
oacdutra, line 150: I'm not sure if this is correct!

tests/web_asset.rs
```diff
@@ -63,7 +63,14 @@ async fn contract_legacy_import() {
...
     let req = serde_wasm_bindgen::to_value(&contract_import).expect("oh no!");
-    let _ = resolve(import_contract(sk, req)).await;
+    let resp = resolve(import_contract(sk.clone(), req)).await;
+    let resp: ContractResponse = json_parse(&resp);
+
+    // TODO: This not working =(
+    // let resp: JsValue = resolve(list_contracts(sk)).await;
+    // let resp: ContractsResponse = json_parse(&resp);
+
+    // assert_eq!(1, resp.contracts.len());
```
oacdutra: I believe the problem is when using WASM32 directly, so it doesn't.

If we look at the carbonado.rs file, we notice that according to the compilation, the network prefix is ​​added.

src/carbonado.rs
cryptoquick: I'm not sure that is either, since these aren't actually called by the server as far as I know, and even if they were, the name is provided by the client, which is where we accounted for this. I actually don't really like these conditionally-compiled methods, from what I understand, they're distracting and unused.

src/bin/bitmaskd.rs
```diff
@@ -270,7 +270,10 @@ async fn co_store(
         .open(&filepath)
     {
         Ok(file) => {
-            let present_header = carbonado::file::Header::try_from(&file)?;
+            let present_header = match carbonado::file::Header::try_from(&file) {
+                Ok(header) => header,
+                _ => carbonado::file::Header::try_from(&body)?,
+            };
```
cryptoquick, line 277: Very nice find, thank you!

src/structs.rs
```diff
@@ -237,8 +237,11 @@ pub struct GenesisFormats {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub enum ContractType {
+    #[serde(rename = "contract")]
     Contract = 9,
+    #[serde(rename = "rgb20")]
     RGB20 = 20,
+    #[serde(rename = "rgb21")]
```
cryptoquick, line 244: Good thinking!

cryptoquick (approved):
ACK
Great work, thanks!

josediegorobles (approved):

PR #235, opened 2023-06-08
Watcher and prefetch improviments

oacdutra:
This PR intent some improviments in watchers and prefetch:

- Reduce number of request to the block explorer
- Allow monitoring specific UTXO/Address
- Add tests to list contract from WASM32 target
- Ignore redundant tests


josediegorobles (approved):
ACK

Issue #151, opened 2023-01-31
Debounce synchronize_wallet calls


cryptoquick:
When multiple calls are made to synchronize_wallet, do nothing if they're made too soon. Also make this a configurable amount.

PR #245, opened 2023-06-14
Separate wallet syncing

### LINKED ISSUE #244: Separate wallet sync
We should use a separate method to sync BDK wallets, and remove all sync calls in our current methods. This will reduce the strain we put on Esplora servers.

cryptoquick:
Resolves #244 and #151 

Removes `synchronize_wallet`, and ntroduces two separate methods, `sync_wallet` and `sync_wallets`. The first is only for internal use, and can be passed a single wallet, whereas the second will sync all wallets and can be called from the browser side.

In the future we could add even more fine-grained wallet syncing for a specific descriptor hash, used as a key in #242

### DIFF
--- a/src/bitcoin.rs
+++ b/src/bitcoin.rs
@@ -68,11 +68,9 @@ pub fn decrypt_wallet(
     hash: &SecretString,
     encrypted_descriptors: &SecretString,
 ) -> Result<DecryptedWalletData> {
-    // let hash: &str = hash.0.as_ref();
     let mut shared_key: [u8; 32] = hex::decode(&hash.0)?
         .try_into()
         .expect("hash is of fixed size");
-    // let encrypted_descriptors: &str = encrypted_descriptors.0.as_ref();
     let encrypted_descriptors: Vec<u8> = hex::decode(&encrypted_descriptors.0)?;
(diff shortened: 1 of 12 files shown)


oacdutra:
Hi, 

I need to point out two things:

1. I added [this](https://github.com/diba-io/bitmask-core/pull/245/files#diff-2d87d39601d1f86ca975fde1a8f9187ec76259696aecfb0f208209b409b8d7e5R185), because we need fill the wallet with data. This is correct?
2. I worked in similar implementation, see [that](https://github.com/diba-io/bitmask-core/blob/development/src/rgb/wallet.rs#L151-L167). I did it because of resolvers and prefetches. I always get the first 20 as that is more than enough for RGB, given that every transition in RGB adds a tweak to the output script, so "we never reuse an address". We could also study using it in Bitcoin, then remove the BDK if there is no longer a need.

cryptoquick:
I still need to do two more things:

We need a good way to make sure all the correct descriptor strings are loaded into the BDK global wallets BTreeMap. We can't just put the result of DecryptedDescriptors into that naively, since not all of those are used.

There's also a bug I found in get_wallet that I'm too tired to explain.

Will investigate more and address these tomorrow.

josediegorobles (approved):
ACK! That's a very good work

PR #252, opened 2023-06-27
Consecutive Transfer

### LINKED ISSUE #17: Transaction fee cost
Get the transaction fee before initiating the transaction, using a broadcast option on send_sats.

### LINKED ISSUE #37: Keep track of the lost sats
In send_tokens we are sending to our own address 546 sats but the public key is being tweak and the wallet don't know that exists. So we are burning a bit of sats as a second "fee". We have to correct this and get the utxo in our wallet.

### LINKED ISSUE #140: Consecutive transfers

### LINKED ISSUE #181: Additional RGB transfer integration tests

oacdutra:
### **Description**

This PR allow users make consecutive transfers using tapret1st method.

**New Features**
* Spend an UTXO with Taproot Tweak
* Transfer a change RGB Contract

**Improviments**
* Bump for RGB version v0.10.4
* Discovery "Tweaked Output" with Watcher Inspect
* Fee Estimate to Descriptor Wallet
* Refactor Integration Tests
* Display if RGB Allocation has been spent

**Bug Fix**
* Missing Precision information 
* Retrieve Supply by Global State

**Tests Scenarios (transfers with check balances)**
- [x] An issuer send a transfer to a new owner.
- [x] An issuer make a transfer with a change amount to a new owner.
- [x] An issuer make a consecutive transfers with a change amount to a new owner.
- [x] An owner send a transfer to another owner. 
- [x] Issuer and owner send two transfers to another owner with different UTXOs.
- [x] Issuer and owner send two transfers to another owner with same UTXOs. 
- [x] Owner use two or more UTXOs to send a transfer to another owner.

**Limitations**
* The user can only generate an invoice if he previously imports the contract (genesis).
* It is not possible to test UDAs per hour, as we can only issue 1 at a time.
* Multiple PSBT inputs must be of the same descriptor.
* Fee is calculated automatically only if not sent in `PsbtRequest` (This is still experimental).

**Break Compatibility [WARNING]**
As mentioned here, this version breaks compatibility with previous releases.

PS:
@trevor-ofarrell and @josediegorobles,
If you have any questions about the structs changes, please contact me.

Closes #17 
Closes #37 
Closes #140  
Closes #181


josediegorobles (approved):
ACK, but now we have this problem with the tests

josediegorobles (approved):

src/structs.rs
```diff
@@ -272,7 +285,7 @@ pub struct ContractResponse {
     /// Amount of the asset
     pub supply: u64,
     /// Precision of the asset
-    pub precision: u64,
+    pub precision: u8,
```
cryptoquick, line 288: Good catch!

cryptoquick (approved):
I can't yet test this due to the error here: https://github.com/diba-io/bitmask-core/issues/263

That said, we can fix that in a bugfix branch, and since this has so many other changes, we should really get it into the development branch so other work can proceed.

PR #254, opened 2023-06-29
CD/re-issue fixes

cryptoquick:
Adds clippy lints for tests to CI in addition to lib. Fixes linter issues. Also reverses the retrieve argument order.


josediegorobles (approved):
LGTM

PR #255, opened 2023-07-05
Update to Strict-Type 1.4.x

oacdutra:
Description

This PR allows us use the latest version of the Strict-Type 1.4.x.

~I'll keep it as a draft because I found a problem in this version: the encoding and decoding operations are 10-20 times slower than the previous one. I believe it is a flaw in the way of decoding text in ASCII format.~

Solved: https://github.com/strict-types/strict-types/issues/21


josediegorobles (approved):
LGTM

cryptoquick (approved):
ACK

PR #267, opened 2023-07-12
Fix Payjoin: Add back original PSBT input to payjoin proposal

DanGould:
Unlike Bitcoin Core's walletprocesspsbt RPC, BDK's finalize_psbt only checks if the script in the PSBT input map matches the descriptor and does not check whether it has control of the OutPoint specified in the unsigned_tx's TxIn. BIP 78 spec clears script data from payjoin proposal. So the original_psbt input data needs to be added back into payjoin_psbt without overwriting receiver input.

See where [we ran into the same problem with BDK-CLI](https://github.com/bitcoindevkit/bdk-cli/pull/156#discussion_r1251172981) and [the proposed solution there](https://github.com/bitcoindevkit/bdk-cli/pull/156#discussion_r1261384934).

### DIFF
--- a/src/bitcoin/payment.rs
+++ b/src/bitcoin/payment.rs
@@ -5,10 +5,10 @@ use bdk::{wallet::tx_builder::TxOrdering, FeeRate, TransactionDetails};
 use bitcoin::{
     consensus::serialize,
     psbt::{Input, Psbt},
-    TxIn,
+    Amount, TxIn,
 };
 
-use payjoin::{PjUri, PjUriExt};
+use payjoin::{send::Configuration, PjUri, PjUriExt};
 
 use crate::{
     bitcoin::{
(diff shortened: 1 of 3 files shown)


cryptoquick (approved):
Neat! LGTM, we can test this in our next release.

DanGould:
I've added ~2~ 1 more relatively small commit.

- One bumps the library `0.8.0` (was set to =0.8.0 since `0.8.1` updates to rust bitcoin 0.30.~ oops! I yanked 0.8.1)
- Second makes use of the fee parameter. PDK needs an upgrade to the fee calculation and estimation, but I implemented it manually here.

src/bitcoin/payment.rs
```diff
@@ -57,9 +64,14 @@ pub async fn create_payjoin(
...
     info!("Original PSBT successfully signed");
 
-    // TODO use fee_rate
-    let pj_params = payjoin::sender::Configuration::non_incentivizing();
-    let (req, ctx) = pj_uri.create_pj_request(original_psbt, pj_params)?;
+    let pj_params = match fee_rate {
+        Some(fee_rate) => {
+            payjoin::send::Configuration::min_fee_rate_sat_per_vb(self, fee_rate.as_sat_per_vb())
+        }
+        None => payjoin::send::Configuration::non_incentivizing(),
```
DanGould: Some(fee_rate) will enforce that the receiver provides enough fees to at least match the sender's preferred feerate. It doesn't yet allow the receiver to spend some from sender's change to cover that feerate (which imo it could for up to 1 additional input since that is a worthwhile spend for privacy). None sets no such requirement nor offers a fee budget to the receiver.

cryptoquick:
These changes still look fine. If any fixes need to be made, we can do them in a separate PR. I'll be sure to run CI checks on this as well.

cryptoquick:
I've invited you to the project with write access so your contributions can run in CI more easily.
For now I'm running CI checks for this PR here:
https://github.com/diba-io/bitmask-core/actions/runs/5545431781

cryptoquick:
Looks like there was a linter issue. I'm closing this for #269.

PR #269, opened 2023-07-13
Fix Payjoin: Add back original PSBT input to payjoin proposal

cryptoquick:
Fixes #267

### DIFF
--- a/src/bitcoin/payment.rs
+++ b/src/bitcoin/payment.rs
@@ -5,10 +5,10 @@ use bdk::{wallet::tx_builder::TxOrdering, FeeRate, TransactionDetails};
 use bitcoin::{
     consensus::serialize,
     psbt::{Input, Psbt},
-    TxIn,
+    Amount, TxIn,
 };
 
-use payjoin::{PjUri, PjUriExt};
+use payjoin::{send::Configuration, PjUri, PjUriExt};
 
 use crate::{
     bitcoin::{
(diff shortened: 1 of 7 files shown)


DanGould:
Thanks for this. As you can see I did 3 things with this PR (separated by commit of course). Is this best practice for this repo or would you prefer separate PRs for distinct features in the future?

cryptoquick:
You'll see we have our repo configured so that it's all squashed when it's merged, and the PR can be referenced instead:
https://github.com/diba-io/bitmask-core/commits/development

src/bitcoin/payment.rs
```diff
@@ -88,3 +128,34 @@ pub async fn create_payjoin(
...
+        )
+    }
+
+    let mut original_inputs = input_pairs(original_psbt).peekable();
+
+    for (proposed_txin, mut proposed_psbtin) in input_pairs(&payjoin_psbt) {
+        if let Some((original_txin, original_psbtin)) = original_inputs.peek() {
+            if proposed_txin.previous_output == original_txin.previous_output {
+                proposed_psbtin.witness_utxo = original_psbtin.witness_utxo.clone();
+                proposed_psbtin.non_witness_utxo = original_psbtin.non_witness_utxo.clone();
```
DanGould, line 155: This is broken because propose_psbtin is cloned, so this does not mutate the original_psbt as it needs to. PR inbound

PR #270, opened 2023-07-13
Add an RGB_LIB_IDs.toml file, and build script to auto-generate it

cryptoquick:
This helps us associate RGB semantic IDs to help us track consensus-breaking changes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,61 @@
+use std::{collections::BTreeMap, fs};
+
+use anyhow::Result;
+use rgbstd::{
+    interface::{LIB_ID_RGB20, LIB_ID_RGB21, LIB_ID_RGB25},
+    stl::{LIB_ID_RGB_CONTRACT, LIB_ID_RGB_STD},
+};
+use serde::{Deserialize, Serialize};
+
+type IdVersionMap = BTreeMap<String, String>;
+
+#[allow(non_snake_case)]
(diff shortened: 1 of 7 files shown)


oacdutra (approved):

oacdutra (approved):

build.rs
```diff
@@ -0,0 +1,67 @@
+use std::{collections::BTreeMap, fs};
+
+use anyhow::Result;
+use rgbstd::{
+    interface::{LIB_ID_RGB20, LIB_ID_RGB21, LIB_ID_RGB25},
+    stl::{LIB_ID_RGB, LIB_ID_RGB_CONTRACT, LIB_ID_RGB_STD},
```
dr-orlovsky, line 6: `LIB_ID_RGB_STD` is not about consensus: if it changes it means you may need to update consignments or indexes or serialization for stash - but not the fact that you need to re-issue and migrate the assets

`LIB_ID_RGB20, LIB_ID_RGB21, LIB_ID_RGB25` - this will mean you will need to do another interface implementations (which is trivial) and doesn't need to re-issue or migrate the assets

`LIB_ID_RGB` - this is the only consensus thing which (if changes) means assets have to be re-issued

oacdutra: Thanks to clarify @dr-orlovsky

PR #273, opened 2023-07-15
Support to Dustless Transactions

### LINKED ISSUE #191: Dustless transfers
For RGB privacy, token transfers should not depend upon dust outputs. The problem with dust outputs is, they're more visible on-chain, and easier to censor.

We have three wallets, one for onchain payments, one for fungible assets (for settlement with token channels), and one for unique digital assets (which are always on-chain).

To pay for fees, we'll transfer sats in and out of these different wallets. If on-chain sats are needed from outputs using tokens, that output is spent, and the tokens on it are transferred to the change output and sent back to the tokens wallet, as a part of a sats spend.

With every token transfer, two outputs are made with random, non-dust amounts. The only issue I can imagine here is output association, and I'm not sure of a good way to address this concern, but it's at least better for on-chain privacy than using dust outputs.

This results in what's essentially a sort of coinjoin behavior in that it might look like you're making payments to others when you're actually just sending sats back to yourself. And if you were sending sats to others, so long as you weren't using rounded payment amounts, they'd look the same, so it'll always be ambiguous if you're sending sats to others or if you're just using RGB tokens.

CC @DanGould, in case payjoin might be useful in this context.

I'm assigning this to the 0.7 release since this could be complex to do correctly. This should still be compatible with spending older 0.6 dust outputs.

oacdutra:
**Description**
This PR allow construct and sign psbt with multiple inputs with different terminal derivations. 

**Changes (WASM)**
Some structs was changes to support this feature, please see:
- PsbtRequest
- PsbtInputRequest
- SignPsbtRequest

Closes #191


tests/rgb/unit/stl.rs
```diff
@@ -7,35 +7,35 @@ use rgbstd::{
 
 #[tokio::test]
 async fn check_rgb20_stl() -> Result<()> {
-    let schema = rgb20_stl();
-    assert_eq!(LIB_ID_RGB20, schema.id().to_string());
+    let shema = rgb20_stl();
+    assert_eq!(LIB_ID_RGB20, shema.id().to_string());
```
cryptoquick, line 11: Why is this being put back? Was schema not misspelled?

tests/rgb/web/stl_ids.rs
```diff
@@ -10,30 +10,30 @@ wasm_bindgen_test_configure!(run_in_browser);
 
 #[wasm_bindgen_test]
 async fn check_rgb20_stl() {
-    let schema = rgb20_stl();
-    assert_eq!(LIB_ID_RGB20, schema.id().to_string());
+    let shema = rgb20_stl();
+    assert_eq!(LIB_ID_RGB20, shema.id().to_string());
```
cryptoquick, line 14: See also here

cryptoquick (changes requested):
These all look like really good changes! Just one small nit.

tests/rgb/unit/stl.rs
oacdutra: Merge problem, probably.

i will fix it, ok?

cryptoquick: Okay!

cryptoquick (approved):
ACK

PR #276, opened 2023-07-17
Add LIB_ID_RGB to version stock

cryptoquick:
Two questions for you, @crisdut:

1. I'm not sure why I get this error in `carbonado.rs`:
```
unresolved import `rgbstd::stl::LIB_ID_RGB`
no `LIB_ID_RGB` in `stl`
```
2. I'm not sure if we also need to do versioning in the rgb_wallets, and what consequence this could have. It would be nice to keep a list of old contracts, but if we don't have the data for them, that could probably cause a panic, so that will need to be tested and handled.


cryptoquick:
Apparently, `#1` is only an issue for me locally. Not sure why.

PR #279, opened 2023-07-18
Carbonado Alternative Names

oacdutra:
**Description**

To avoid overwriting old carbonado files, we started to adopt the new pattern and prepare the carbonado module to verify and copy old formats to new filename format.

Also, we check only carbonado files was metadata (to avoid break the server if oldest version of the strict-types).

**Old Patterns (Check):**
- {filename}
- {network}-{filename}
- {network}/{filename}
- {network}-{rgb_id_version}-{filename}

**New Pattern**
- {network}/{rgb_id_version}/{filename}


josediegorobles (approved):
LGTM

PR #280, opened 2023-07-18
Maintain and enforce percent-encoded filenames on-disk.

cryptoquick:


src/carbonado.rs
```diff
@@ -321,28 +321,33 @@ pub async fn retrieve(
...
 
     let network = NETWORK.read().await.to_string();
     let mut final_name = name.to_string();
     if !name.contains(&network) {
         final_name = format!("{network}/{name}");
     }
 
-    let directory = std::path::Path::new(
+    let reencoded = utf8_percent_encode(&percent_decode_str(name).decode_utf8()?, FORM).to_string();
+    if name != reencoded {
```
oacdutra, line 334: I'm not sure if it works.

Because to pass the file path, we need to run the url encode.

Ex (WASM32):

This path params `filename` works because we encode the "`/`" to "`V%2F`"

http://.../carbonado/02...e/regtest/**memphis_asia_crash`V%2F`bitmask-fungible_assets_stock.c15**

But if we send without encode, the server understand the "/" is a path and return 404.

http://.../carbonado/02...e/regtest/**memphis_asia_crash`/`bitmask-fungible_assets_stock.c15**

src/carbonado.rs
```diff
@@ -321,28 +321,33 @@ pub async fn retrieve(
...
+    let directory: PathBuf = std::path::Path::new(
         &std::env::var("CARBONADO_DIR").unwrap_or("/tmp/bitmaskd/carbonado".to_owned()),
     )
     .join(pk);
-
-    let final_name = percent_decode(final_name.as_bytes()).decode_utf8().unwrap();
-    let filepath = directory.join(final_name.to_string());
-    let filedir = filepath.parent().unwrap();
-    fs::create_dir_all(filedir).await.map_err(|_| {
+    fs::create_dir_all(&directory).await.map_err(|_| {
```
oacdutra, line 342: Similar to describe above.

```suggestion
    let final_name = percent_decode(final_name.as_bytes()).decode_utf8().unwrap();
    let filepath = directory.join(final_name.to_string());
    let filedir = filepath.parent().unwrap();
    fs::create_dir_all(filedir).await.map_err(|_| {
```

src/carbonado/error.rs
```diff
@@ -5,6 +5,8 @@ use thiserror::Error;
 pub enum CarbonadoError {
     /// std io error
     StdIoError(#[from] std::io::Error),
+    /// std str UTF-8 error
+    StdStrUtf8Error(#[from] std::str::Utf8Error),
```
oacdutra, line 9: Nice!

oacdutra (changes requested):

PR #282, opened 2023-07-18
Use a single transaction to fund vault

cryptoquick:
This might help make funding the vault work on mainnet, since we're not spending a transaction immediately after making it.


josediegorobles (approved):
LGTM

PR #286, opened 2023-07-19
Fix transfer tests

oacdutra:
Alternative #284


src/rgb/contract.rs
```diff
@@ -314,7 +306,7 @@ where
     }
 
     let resp = ContractResponse {
-        contract_id,
+        contract_id: contr_id,
```
cryptoquick, line 309: Can you put this back?

cryptoquick (changes requested):

cryptoquick:
Good work so far, but it would help if you rename it back to contract_id.

cryptoquick (approved):
ACK

PR #288, opened 2023-07-20
Fix MimeType and Attach

oacdutra:
**Description**

This PR add two changes:
- Fix mime types
- Add attachment data: 
image

Also, I needed change the struct to support that:

image

RGB-Wallet PR:
https://github.com/RGB-WG/rgb-wallet/pull/89


cryptoquick (approved):
Tested ACK

PR #293, opened 2023-07-24
Release BMC v0.6-rc17

### LINKED ISSUE #295: Return Genesis Timestamp in RGB Struct
I'm pretty sure RGB embeds a timestamp in contract genesis, or at least, it used to. We need that.

oacdutra:
**Description**
- Bump rgb-wallet to v0.10.4
- Bump strict-ytpes to v1.6

PS: After approved, we need update the production server.

Closes #295


cryptoquick (approved):
ACK

PR #298, opened 2023-07-25
Nostr keys fix

cryptoquick:
Turns out, how we generate nostr keys is broken, after comparing to an online NIP-06 generator.

- [x] Tests
- [x] Fix


src/bitcoin/keys.rs
```diff
@@ -105,27 +109,19 @@ fn watcher_xpub(
...
 }
 
-fn nostr_keypair(
-    xprv: &ExtendedPrivKey,
-    path: &str,
-    change: u32,
-) -> Result<(String, String), BitcoinKeysError> {
+// For NIP-06 Nostr signing and Carbonado encryption key derivation
+fn nostr_keypair(xprv: &ExtendedPrivKey) -> Result<(String, String), BitcoinKeysError> {
+    pub const NOSTR_PATH: &str = "m/44'/1237'/0'/0/0";
```
oacdutra, line 114: According with slip44, we need change account in testnet networks.

[BIP32](https://bips.xyz/32) is used to derive the path m/44'/1237'/<account>'/0/0 (according to the Nostr entry on [SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).

```suggestion
    pub const NOSTR_PATH: &str = format!("m/44'/1237'/{network_account}'/0/0");
```

tests/keys.rs
```diff
@@ -0,0 +1,75 @@
...
+        "correct nsec is generated"
+    );
+    assert_eq!(
+        wallet_data.public.nostr_npub,
+        "npub1v5zwxyd3dtmrvgnamxxlfxj92t52hwkg09dmwjhmkjujkq8kdzms77547c",
+        "correct npub is generated"
+    );
+
+    Ok(())
+}
```
oacdutra, line 75: We could create more descriptor related tests on networks like testnet for example.

oacdutra (changes requested):

src/bitcoin/keys.rs
cryptoquick: The only reason I left that out was because we don't have a way to specify that in the UI, and no way to pass an account argument in any of our key generation. That's a whole separate feature, it has not yet been requested, and I'm not sure how well it would work in our UI. It's hard enough to support different networks. I'll create a low priority tracking issue for it, however.

tests/keys.rs
cryptoquick: Mainnet is the thing I'm most worried about. Anything else is overkill, especially since a lot of that is handled automatically by rust-bitcoin / BDK.

src/bitcoin/keys.rs
oacdutra: > The only reason I left that out was because we don't have a way to specify that in the UI, and no way to pass an account argument in any of our key generation. That's a whole separate feature, it has not yet been requested, and I'm not sure how well it would work in our UI. It's hard enough to support different networks. I'll create a low priority tracking issue for it, however.

This argument can be used when switching networks via UI, where bitcoin=0 and others=1.

Without this change, Nostr's keys will always be the same.

tests/keys.rs
oacdutra: > Mainnet is the thing I'm most worried about. Anything else is overkill, especially since a lot of that is handled automatically by rust-bitcoin / BDK.

See previous comment above!

src/bitcoin/keys.rs
cryptoquick: No, it's not network account, they're separate; network is 1237, and account is 0. I don't think it makes sense to give users a different account when they're on testnet. Remember, this will give them different Nostr keys, also.

oacdutra: Oh sorry, I made a mistake. I will approve your PR.

oacdutra (approved):

src/bitcoin/keys.rs
cryptoquick: Haha, no worries, I appreciate the thorough review!

PR #305, opened 2023-07-28
Update swap_btc_ln() function

grunch:
Now we can set the lightning address where the user will receive the sats from the swap

### DIFF
--- a/src/lightning.rs
+++ b/src/lightning.rs
@@ -254,9 +254,13 @@ pub async fn check_payment(payment_hash: &str) -> Result<bool> {
 }
 
 /// Swap BTC onchain to Lightning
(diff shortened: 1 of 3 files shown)


cryptoquick (approved):
ACK

Will get this out in a new release soon

josediegorobles (approved):
ACK

cryptoquick (changes requested):
There's an error in the `create_wallet_test` test now, and I'm not sure why.
I've updated the assertion to assert_eq so it's easier to debug. You can test it locally with `cargo test create_wallet_test`.

grunch:
I re-ran tests and I got `create_wallet_test` passed

cryptoquick:
Awesome

cryptoquick (approved):
Approved

PR #309, opened 2023-07-31
Find additionalfeeoutputindex from unsigned_psbt

DanGould:
outputs map does not contain the script_pubkey, usigned_tx txos do. PSBTv2 NOW!


cryptoquick (approved):
Oh nice, we'll have to see if this works in our next release

PR #311, opened 2023-08-01
Validate nostr keys

cryptoquick:
Validate nostr keys and trim them if slightly too long (33 byte x-only pk).

This way the key from `vault.public.nostr_pub` can be provided directly, otherwise, the way we generate it, it's just a little too long. We generate it this way because the extra byte is helpful for Carbonado encryption.


grunch (approved):
LGTM

cryptoquick:
ty!

PR #315, opened 2023-08-02
Automatically convert npub into hex

### LINKED ISSUE #314: Parse npubs in nostr methods
Currently we only support hex input, but we do already use NDK, so we might as well parse them out into their hex equivalents automatically.

cryptoquick:
Closes #314


josediegorobles (approved):
LGTM

PR #316, opened 2023-08-02
fix genesis import and reduce esplora calls

oacdutra:


src/rgb/prefetch.rs
```diff
@@ -395,11 +401,11 @@ pub async fn prefetch_resolver_waddress(
     let esplora_client: EsploraBlockchain =
         EsploraBlockchain::new(&explorer.explorer_url, 1).with_concurrency(6);
 
+    let index = 0;
     let mut step = 100;
```
cryptoquick, line 405: Does this line also need to be like these lines?
https://github.com/diba-io/bitmask-core/pull/316/files#diff-16d56c26391d31365bd04f751b5c0e1ac501850ceb89e167b5a96931c0e5037bR302-R305

josediegorobles (approved):
LGTM

cryptoquick (approved):
ACK

PR #317, opened 2023-08-03
fix: blank transition for different contract types

oacdutra:
**Description**

This PR has an interim fix that allows you to create transitions for all contracts linked to the single UTXO, regardless of its implementation (iface).


josediegorobles (commented):
ACK

PR #319, opened 2023-08-04
Fix watcher next utxos

oacdutra:
**Description**

This PR fix the method **watcher_next_utxo**, avoiding duplication of the same information and listing UTXOs containing anchored contracts (previously, we prohibited this due to the problem of consecutive transfers, but this filter went unnoticed after the correction).a


josediegorobles (approved):
LGTM

PR #322, opened 2023-08-05
Storage Transfers (Consig) Files

### LINKED ISSUE #318: Temporary Stock
We need to have a temporary stock for pending transfers before they're added to the user's stock. It can also help prevent adding failed transfers to the user's stock. This can then be cleared by replacing it with an empty array if need be.

### LINKED ISSUE #321: Support to coin selection (bitcoin) in Full Transfer Asset

oacdutra:
### **Description**

This PR intent add support to storage consig files temporally to allow users waiting the transaction is publish and confirmed in blockchain.

**New Features**
We add three new methods:

* **list_transfers:**  Allow us list all transfers related with a contract (RgbTransferDetail).
* **save_transfer:** Allow us save transfer related with a contract.
* **remove_transfer:** Allow us remove one or more transfer(s) related with a contract.

**Changes**
* Transfer operation save consig file automatically to the sender.

**Fixies**
* Resolver and Prefetch for UTXOs now check all vouts.

**RgbTransferDetail (struct)**
image

Closes https://github.com/diba-io/bitmask-core/issues/318
Closes #321

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ futures = { version = "0.3.28", features = [
     "executor",
 ], default-features = true }
 garde = { version = "0.11.2", features = ["derive"], default-features = false }
+rand = "0.8.5"
 getrandom = { version = "0.2.10", features = ["js"] }
 hex = "0.4.3"
 indexmap = "1.9.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bitmask-core"
-version = "0.6.3-rc.1"
+version = "0.6.3-rc.2"
 authors = [
     "Jose Diego Robles <jose@diba.io>",
     "Hunter Trujillo <hunter@diba.io>",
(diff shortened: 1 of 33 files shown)


josediegorobles (approved):
ACK

oacdutra:
@josediegorobles There is an example how your which parameters you need to use in your call:
https://github.com/diba-io/bitmask-core/blob/95c797b68780ca58d852c949cec6f27606752cb8/tests/rgb/integration/internal.rs#L134-L143

Cargo.toml
```diff
@@ -51,6 +51,7 @@ futures = { version = "0.3.28", features = [
     "executor",
 ], default-features = true }
 garde = { version = "0.11.2", features = ["derive"], default-features = false }
+rand = "0.8.5"
```
cryptoquick, line 54: I think this is unused, and even if it was used, we use getrandom now.

cryptoquick (changes requested):
Nice work! Just one change requested.

Cargo.toml
oacdutra: We use that to generate a random number between two numbers.

I think getrandom does not provide that, right?

cryptoquick: I see... Well, we can keep that, then!

cryptoquick (approved):
ACK

PR #326, opened 2023-08-07
Safe watcher migration

oacdutra:
**Description**

This PR allow us verifies that the xpub within the current wallet is correct. If not, move the old wallet to "legacy" and register the new wallet.


josediegorobles (approved):
LGTM

cryptoquick (approved):
Wow, that's actually pretty neat!

PR #328, opened 2023-08-08
Fix prefetch utxos filters

oacdutra:
**Description**

This PR fixes the spent UTXO filter for wasm32 compilation.

Related #319


josediegorobles (approved):
Tested ACK

PR #332, opened 2023-08-11
fix: add tapret in full-transfer operation

oacdutra:
**Description**

This PR detect and add tapret commits automatically in **full_transfer** operation


josediegorobles (approved):
Tested ACK

PR #335, opened 2023-08-14
Adding iface to batch response

josediegorobles:


oacdutra (commented):
We can change the saveTransfer to add iface.

That way we would avoid having to call the wallets and the information would already be related in the creation of the same.

cryptoquick:
@crisdut Can you make a PR into #336 adding that? Then I can do an npm rc with both.

josediegorobles:
> We can change the saveTransfer to add iface.
> 
> That way we would avoid having to call the wallets and the information would already be related in the creation of the same.

That's much better

PR #337, opened 2023-08-14
Test Suite for Wasm32 

oacdutra:
**Description**

For help troubleshooting in wasm32 target, I add support to regtest operations in wasm32 target.

**How this works**

Now, when we up the bitmaskd server if "regtest" flag, the server enables two new routes:

- **/regtest/block** Generate a new block in bitcoin node regtest
- **/regtest/block/:address/:amount**  Send an amount of satohis to bitcoin address

This allows us to simulate use cases of the web-wallet and extensions wallet.

For more details, see the `tests/rgb/web/transfers.rs` file.

Also, this PR solves all issues related to transfers between web-wallets.


src/bin/bitmaskd.rs
```diff
@@ -579,12 +597,19 @@ async fn main() -> Result<()> {
...
         .route("/carbonado/:pk/:name/force", post(co_force_store))
         .route("/carbonado/:pk/:name/metadata", get(co_metadata))
-        .route("/carbonado/:pk/:name", get(co_retrieve))
-        .layer(CorsLayer::permissive());
+        .route("/carbonado/:pk/:name", get(co_retrieve));
 
     let network = get_network().await;
     switch_network(&network).await?;
 
+    #[cfg(not(target_arch = "wasm32"))]
```
cryptoquick: I'm not sure this block makes sense, since this file will never be compiled to wasm

cryptoquick (changes requested):
Just one issue, otherwise, this looks really good. Does this also include some bugfixes?

oacdutra:
> . Does this also include some bugfixes?

Yes, I worked closely with @josediegorobles on some issues related to transfers on the wasm32 destination:

1. `bad-txns-inputs-missingorspent` The error was occurred because prefetch is not defined correctly in full_transfer method. (fixed)
2. `The script is invalid: derived scriptPubkey does not match transaction scriptPubkey` The error occurs when you try spend contract with more than one input, each with tweak (tapret). (fixed)
3. Some issues related BME

Also, Thanks to the new endpoints, I was able to reproduce all the steps in BME on BMC, as you can see here:
https://github.com/diba-io/bitmask-core/pull/337/files#diff-f110b43ee23dce915a3b9f8a78373477bdc05c4cfa6dc8c79544c0d7825cfb53R1-R434

cryptoquick (approved):
ACK

josediegorobles (approved):
Tested ACK

PR #341, opened 2023-08-15
Fix: remove incremental compilation

oacdutra:


cryptoquick:
This doesn't seem to solve the issue, does it?

oacdutra:
> This doesn't seem to solve the issue, does it?

No, i think the problem is related with cache. I will create an issue to optimize and organize the CI, ok?

PR #345, opened 2023-08-17
Refactor prefetches utxos 

### LINKED ISSUE #340: Getting current utxo
I have this in the extension:

```
const { utxo } = await backend.rgb.watcherNextUtxo(
    vault.private.nostrPrv,
    "default",
    iface
  );
  let { utxos } = await backend.rgb.watcherUnspentUtxos(
    vault.private.nostrPrv,
    "default",
    iface
  );
if (utxo) utxos = utxos.filter((u) => u.outpoint === utxo.outpoint);
  const seal = `${sealPrefix}:${utxos[0].outpoint}`;
```

I want to get the utxo that is currently using for store assets. I'm calling two methods that is going to be too much (and too much calls to the mempool). I need a single method, watcher_current_utxo with same parameters and returns as watcher_next_utxo

oacdutra:
**Description**
This PR intent refactor prefetchs related with utxos, break into two:
- **Prefetch User Utxos:** This method will be used to load user-related UTXOs, in ascending order.
- **Prefetch Contract Allocations Utxos:** This method will be used to load contract-related UTXOs, to check each state of allocations, without relying on the user's wallet..

Also, now we only accept RGB transactions after the L1 transaction is in the block. This measure aims to protect the stash from eventual inconsistencies.

Closes #340


josediegorobles (approved):
ACK

cryptoquick (approved):
A good solution!

ACK

PR #349, opened 2023-08-19
Fix address reuse in fund_vault

cryptoquick:
This also uses random amounts in the fund_vault transaction. Even though this provides an unusual 4-5 output transaction, with 4 of the outputs dust of different sizes, it should be slightly more resistant to chain analysis in the very least.

This also fixes an issue where the use of a fixed seed RNG would produce the same amount each time it was called.

Notice this changes the fund_vault signature and will need some changes in the extension.


josediegorobles (approved):
ACK

cryptoquick:
@josediegorobles I just realized, how will we get a second UDA and asset address? Maybe @crisdut might have some good ideas for that, since this makes use of RGB wallets, and there's  a get_address method, though I don't think it provides unused addresses, only addresses at a specific index.

josediegorobles:
> @josediegorobles I just realized, how will we get a second UDA and asset address? Maybe @crisdut might have some good ideas for that, since this makes use of RGB wallets, and there's a get_address method, though I don't think it provides unused addresses, only addresses at a specific index.

Maybe get_new_address can get the job done

cryptoquick:
Maybe! I looked into what we use:
https://github.com/diba-io/bitmask-core/blob/42107c3b527d798fb425bfd2c8fc73932934d6a7/src/bitcoin.rs#L301

And it looks like that does increment the index:
https://docs.rs/bdk/latest/bdk/wallet/enum.AddressIndex.html#variant.New

And since we retain wallets in memory now, that should persist when it didn't before. Calling it twice might just work, and if it doesn't, it'd be no different than the existing behavior.

That's a good point, let's try it out.

Issue #268, opened 2023-07-13
RGB Stock CRDTs


oacdutra:
Today the RGB module works based in unit of work concept, or that, each operation can be valid only when everything works, like transaction in database (commit and rollback operations). This approach is a way to prevent inconsistencies in stash.

However, This is  not prevent stash inconsistencies in scenarios of the concurrent and/or parallel operations.

For example, Suppose we have a contract with intensive usage and by UX decisions, we allow the users to send and accept transactions Concurrently. One of both operations can be overring the state of the other, causing an inconsistent stash.

We could use a queue to prevent that, but this turns our infrastructure more centralized and also represents a single point of failure. 

We can explore a way to turn Stash compatible with the CFDT approach (@cryptoquick's suggestion).

cryptoquick:
I think CRDTs are the way. It would be great to use something like automerge.

cryptoquick:
A CRDT solution based on automerge supersedes the solution in #258.

PR #351, opened 2023-08-20
Fix Storage State in Carbonado

oacdutra:
**Description**

I found a problem with carbonado overwriting file verification when we use wasm32 (see [here](https://github.com/diba-io/bitmask-core/blob/development/src/bin/bitmaskd.rs#L433-L439)).

After running consecutive transfers (more than 100x), the **rgb_wallet** or **stock** ends up having a minimum variation of bytes downwards, which ends up being unsaved, and this causes some issues in contract states ([@josediegorobles](https://github.com/josediegorobles) I believe this causes some inconsistencies in the balances you reported).

This PR still needs to be completed, I need some help to test it.


cryptoquick (approved):
Sure, this should be fine. We really should get an automerge implementation of #268, however, I think that would resolve all these issues.

oacdutra:
> Sure, this should be fine. We really should get an automerge implementation of #268, however, I think that would resolve all these issues.

Yes, we finished that: #353

PR #361, opened 2023-08-29
Add RGB invoice chain validation

### LINKED ISSUE #360: Add chain / network to RGB invoice
From what I can tell, what chain / network the wallet is currently set is not being set nor is it validated on RGB invoices.

We should both set it, since there's a field for it, and validate it upon all methods that accept invoices.

oacdutra:
Closes #360

### DIFF
--- a/src/validators.rs
+++ b/src/validators.rs
@@ -1,7 +1,8 @@
 use std::str::FromStr;
 
-use bp::Txid;
+use bp::{Chain, Txid};
 use miniscript_crate::Descriptor;
+use rgbwallet::RgbInvoice;
 use seals::txout::ExplicitSeal;
 use wallet::hd::{DerivationAccount, DerivationSubpath, UnhardenedIndex};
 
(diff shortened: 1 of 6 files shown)


src/validators.rs
```diff
@@ -43,12 +50,22 @@ impl Default for RGBContext {
     fn default() -> Self {
         Self {
             closed_method: "tapret1st".to_string(),
+            current_network: String::new(),
```
cryptoquick, line 53: Would it make sense to default to this being Option::None, since that's the type on RgbInvoice ?

cryptoquick (changes requested):
Very fast, very good work! Just one question.

josediegorobles (approved):
ACK

src/validators.rs
oacdutra: Yes, we can work with option. I will change.

cryptoquick (approved):
ACK

PR #372, opened 2023-09-12
Minimal secrets passing for nostr keys

cryptoquick:
This is a major refactor, and will probably break tests. I'm making this a separate PR, to not only make it easier to review, but also easier to fix. Please review as soon as you can, @crisdut, thanks.

Also, I realize this kills our capability for multitenancy. I think this is fine, and probably for the better. It makes us better at serverless function as a service applications, which is what WASM is good for, anyway. This is more sandboxed and more secure against secrets leakage, since we do keep all descriptor wallets in the same central data structure.


cryptoquick:
I think I'm going to take a different approach to this.

PR #378, opened 2023-09-28
Adding swap methods to bitmaskd

josediegorobles:


oacdutra (commented):
Are these methods only used by the market?

Does it make sense for the market to request these endpoints by passing the sk instead of invoking the seed on the endpoint?

josediegorobles:
Ok, I have to do some work more before I have this clear

cryptoquick:
Be sure this is updated from what's in `development`

josediegorobles:
> Be sure this is updated from what's in `development`

I resolved the merge conflicts but anyway I convert the PR to draft because I'm using it only to test the marketplace, when the PR of the marketplace is ready, I will put this PR to RFR, until then is not necessary

PR #383, opened 2023-10-03
bitmask-core.js

cryptoquick:
I've added our TypeScript definitions directly to BMC. The reasoning behind this are:

- Discoverability; when a newcomer arrives to our project, they can quickly see a portion of the files are written in TypeScript
- Maintainability; when a web.rs method is changed, a corresponding method in lib/ts should also be changed in the same PR.

This also opens the door to TS / JS examples and tests and workflows and such.


josediegorobles (approved):
ACK

PR #388, opened 2023-10-05
Adding structs for rgb swap code in development

josediegorobles:


cryptoquick (approved):
ACK

Nice! It took me a minute to realize I was looking at TypeScript in my Rust repository. Yes, this is exactly what why we should have them both together, they got out of sync slightly here too.

PR #391, opened 2023-10-14
Fixes for RGB transfer return encodings

cryptoquick:
- [x] Return ASCII-armored Base85 consignments
- [x] Return transfer txid


mahmoudeA:
https://github.com/diba-io/bitmask-core/issues/456

PR #396, opened 2023-10-19
Support to Fractions

### LINKED ISSUE #363: Transfer Fractions
Right now we only transfer amounts in u64 using bigint. However, RGB contracts support decimal precision up to, I think, 255. We need to be able to pass a bigint for the numbers after the decimal point to be used as the fraction, as a separate argument.

- [ ] Fraction Tests
- [ ] Fraction Structs
- [ ] Fraction Transfer
- [ ] Update web.rs

oacdutra:
### **Description**

This PR add support to fractions in issuer, create invoice and transfers (include swaps) operations.

#### **Changes (!!Important!!)**

I was need change some **pre-reqs** of the issue #363 for feature work correctly.  In my tests, I identify some problems with we send only whole and fraction. See bellow:

1. **Conversions:** Each language (Rust, TS and JS) have different ways to convert and handle numbers and this leaves room for countless errors. To avoid wrong conversions, I decided change the type of `contract amount` from number to string and handle in BMC. The BME send the follow representations of the fractional number: `2.`, `2.0`, `2.0`, `2.0000001`, `2.5`, etc..
2. **Precisions:** Another problem is contract precisions. To avoid user send wrong  values in invoice, I change the invoice amount to string too. Also, now the **receiver** needs already imported contract, because we checked if user send a correct precision number. Also, I create the method called **get_simple_contract** in web.rs to return the contract precision data.

Closes #363

### DIFF
--- a/lib/web/rgb.ts
+++ b/lib/web/rgb.ts
@@ -327,6 +327,16 @@ export interface MediaInfo {
   source: string;
 }
 
+// In structs.rs this is called SimpleContractResponse
+export interface SimpleContractResponse {
+  /// The contract id
+  contractId: string;
+  /// The contract interface
+  ifaceId: string;
+  /// Precision of the asset
+  precision: number;
+}
+
 export interface InvoiceRequest {
   /// The contract id
   contractId: string;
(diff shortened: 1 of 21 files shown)


josediegorobles (approved):
ACK

PR #398, opened 2023-10-21
Support to Batch Operations

### LINKED ISSUE #304: RGB Transfer Batching
When a transfer is initiated, sync that specific RGB wallet, and check to see if there's already an RGB transfer operation in progress (unmined), and RBF with a new transfer that consolidates tokens onto the same change UTXO. If it's from the same UTXO, this should be done anyway, but if it's from a different UTXO, it can still save on fees. Combine with Payjoin for best results. If the RBF cannot be performed due to a block being mined at the same time, the transfer is adjusted accordingly.

oacdutra:
**Description**

This PR intent supports to RBF and Batch Operations.

**How to works**

The BMC will always check for pending transitions related to a specific wallet (RGB20, RGB21, etc..). If have, the BMC retrieves the invoices of the previous RGB transitions and will generate a new PSBT and consignment file(s), consolidating all previous changes in new one. Another feature implemented is **strict consignments**. This flag allows us to create consignment files per invoice, so we can send it to each user, who only knows about the state that interests them, without exposing other allocations.

We will use RGB Proxy to control the distribution of new files to users.

~**11/21/23:** This PR is in draft, as the addition of the proxy and integration will be done in another PR, to facilitate review;~

Closes #304


src/rgb/consignmnet.rs
```diff
@@ -0,0 +1,370 @@
+use std::{
```
cryptoquick: This file is misnamed, it should be `consignment.rs`

src/rgb/consignmnet.rs
```diff
@@ -0,0 +1,370 @@
...
+        // Here we assume the provided PSBT is final: its inputs and outputs will not be
+        // modified after calling this method.
+        let bundles = psbt.rgb_bundles()?;
+        // TODO: Make it two-staged, such that PSBT editing will be allowed by other
+        //       participants as required for multiparty protocols like coinjoin.
+        psbt.rgb_bundle_to_lnpbp4()?;
+        let anchor = psbt.dbc_conclude(method)?;
+        // TODO: Ensure that with PSBTv2 we remove flag allowing PSBT modification.
+
+        // 4. Prepare transfer
```
cryptoquick:
```suggestion
        // 5. Prepare transfer
```

src/rgb/prefetch.rs
```diff
@@ -678,3 +728,30 @@ pub async fn prefetch_resolver_txs_status(txids: Vec<Txid>, explorer: &mut Explo
...
+
+#[cfg(target_arch = "wasm32")]
+struct ExploreAsyncExt {}
+
+#[cfg(target_arch = "wasm32")]
+impl ExploreAsyncExt {
+    pub async fn get_full_tx(
+        client: &AsyncClient,
+        txid: &bitcoin::Txid,
+    ) -> Result<ExploraTX, ExploreClientExtError> {
```
cryptoquick: This should be called `ExplorerTx`

src/rgb/psbt.rs
```diff
@@ -473,66 +475,84 @@ where
...
-            input.clone(),
-            wallet.clone(),
-            tx_resolver,
-        )
-        .map_err(|op| PsbtInputError::WrongWatcherTweak(op.to_string()))?
-        {
-            input.tweak = Some((
-                Fingerprint::default(),
-                tweak
+        input.tweak = match psbt_input.tapret {
```
cryptoquick, line 523: Why are we doing our own verification here? Shouldn't we be delegating that to RGB wallet / core?

cryptoquick (changes requested):
Just some nits / questions

cryptoquick (approved):
ACK

PR #399, opened 2023-10-24
Support RGB Proxy

### LINKED ISSUE #370: Iris Compatibility
We need to be compatible with Iris wallet, for both RGB20 and RGB21.
They don't appear to have a version that supports RGB 0.10 out, so we're blocked on this somewhat until that's released, but there's no reason we can't communicate and coordinate on things like attachment formats.

oacdutra:
**Description**

This PR intent adds support to RGB Proxy in BMC, allowing us to share files between users. Today, we share only two file types: consignment and media.

Also, I made changes in the issue process to interop with Iris Wallet.

Closes #370


josediegorobles (approved):
ACK

cryptoquick (approved):
ACK

PR #400, opened 2023-10-25
Consignment proxy support fixes

cryptoquick:
Be sure to update lib/web whenever changing web.rs. Also, I found several typos, and some would have completely broken this if we had gone to implement with it.


josediegorobles (approved):
LGTM

oacdutra (approved):

PR #410, opened 2023-11-07
Change proxy

josediegorobles:


cryptoquick (approved):
ACK

oacdutra:
Hi @josediegorobles 

I think this change does not work because today, when we send consignment or media to the proxy, we use `bitmaskd` to download the file temporarily before sending it to `rgb-proxy`.

cryptoquick:
Good point, nACK

PR #411, opened 2023-11-07
Reduce dust transactions and hide invalid states

### LINKED ISSUE #375: Combine Token Change with Witness Transaction Output
- [ ] Assign Change Tokens to Witness Transaction Output
- [ ] Only need 2 dust UTXOs, one for RGB20 and one for RGB21

Simplifies transfers, reduces fees

oacdutra:
Closes #375 

**Update 11/9**

I found three new bugs in bitmask-core (all fixed):

1. In WASM target, the export contract cannot hide invalid states
2. When we save the tapret commit, we cannot save the utxo amount 
3. The PSBT signal and transmission were not exposed in the typescript

Also, I made a new test to check if we can make a transfer with  invalid states in the stock [1]

[1] RGB invalid states occurs when we generate a new consignment and the transaction was not broadcasted (i.e. invalid transaction, batch/rbf operations, etc...)


cryptoquick (changes requested):
It looks like `FundVaultDetails` struct was never updated...

Screenshot from 2023-11-07 16-13-12

src/bitcoin.rs
```diff
@@ -441,9 +420,9 @@ pub async fn fund_vault(
 
     Ok(FundVaultDetails {
         assets_output: Some(asset_outputs[0].to_owned()),
-        assets_change_output: Some(asset_outputs[1].to_owned()),
-        udas_output: Some(asset_outputs[2].to_owned()),
-        udas_change_output: Some(asset_outputs[3].to_owned()),
+        assets_change_output: None,
+        udas_output: Some(asset_outputs[1].to_owned()),
+        udas_change_output: None,
```
cryptoquick: I think the change outputs should be completely removed from the struct. Do you think that makes sense?
```suggestion
        udas_output: Some(asset_outputs[1].to_owned()),
```
It will need a matching change in typescript. It's just, if we're hardcoding that they're always none, then it makes sense to omit them entirely, right?

cryptoquick (changes requested):
One more thing, sorry

src/bitcoin.rs
oacdutra: Yes, I think make sense.

I will remove.

josediegorobles (approved):
ACK

PR #421, opened 2023-11-18
Adjust decimal places

oacdutra:


josediegorobles:
It's failing in the swaps flow, could you review that?

cryptoquick (changes requested):
It looks like the wasm flow is failing a few important tests, like RGB transfers and RGB swaps. Like Jose pointed out, that might be a real issue.

That said, these are very good improvements! Will merge and deploy as soon as wasm tests look good.

cryptoquick:
@josediegorobles Looks like this is a heisenbug, see this CI run that had nothing to do with RGB code:
https://github.com/diba-io/bitmask-core/actions/runs/6917007100/job/18817683711

The one right before it ran fine:
https://github.com/diba-io/bitmask-core/actions/runs/6907609426/job/18795211121

josediegorobles:
Ok, re running test is ok, but test-wasm no:

test rgb::rgb::web::transfers::create_transfer_with_fee_value ... FAIL

josediegorobles (approved):
ACK

PR #423, opened 2023-11-20
Periodic metrics

cryptoquick:
Collect metrics periodically to a file, and only return that file from the metrics endpoints.


josediegorobles (approved):
ACK

PR #426, opened 2023-11-24
Add metrics delay

cryptoquick:
So it doesn't bog down our server

### DIFF
--- a/src/carbonado/metrics.rs
+++ b/src/carbonado/metrics.rs
@@ -1,8 +1,13 @@
-use std::{collections::BTreeMap, path::Path, time::SystemTime};
+use std::{
+    collections::BTreeMap,
+    path::Path,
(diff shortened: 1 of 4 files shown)


src/carbonado/metrics.rs
```diff
@@ -1,8 +1,13 @@
...
+use std::{
+    collections::BTreeMap,
+    path::Path,
+    time::{Duration as StdDuration, SystemTime},
+};
 
 use anyhow::Result;
 use chrono::{DateTime, Duration, NaiveDate, Utc};
 use serde::{Deserialize, Serialize};
+use tokio::time::sleep;
```
josediegorobles, line 11: What about this lint error?

josediegorobles (approved):
ACK

PR #428, opened 2023-11-29
Fix #424 by iterating original_input on match

### LINKED ISSUE #424: Send Payjoin Fails to Sign occasionally
Payjoin will occasionally Fail to Sign because original input is skipped rather than checked against the next proposed input. This original_input.next() check should be moved inside the curly braces

https://github.com/diba-io/bitmask-core/blob/af40bc05d3e93d4392d4376da0ddff7cede59a77/src/bitcoin/payment.rs#L177C16-L179

DanGould:
The payjoin psbt construction code reintroduces original psbt inputs that the receiver may have lost. It does this by iterating through each proposed input and original input.

This fix only iterates to the next original_input once a proposed input that matches is found, otherwise inputs could be skipped and signing could fail. The outer loop already iterates through proposed inputs. Inputs in the proposal are in the same order as the original as per spec, and just insert receiver inputs so this iteration will always find matches.


cryptoquick (approved):
Thank you for taking care of this for us :pray:

PR #429, opened 2023-12-05
Fix error on source mdia

josediegorobles:


src/structs.rs
```diff
@@ -112,6 +112,41 @@ pub struct IssueAssetRequest {
     pub request: IssueRequest,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+#[serde(rename_all = "camelCase")]
+#[derive(Validate)]
+#[garde(context(RGBContext))]
+pub struct IssuePreRequest {
```
oacdutra: Just for naming convention, similar to FullTransferRequest.

```suggestion
pub struct FullIssueRequest {
```

src/web.rs
```diff
@@ -381,7 +381,24 @@ pub mod rgb {
...
         future_to_promise(async move {
-            let req: IssueRequest = serde_wasm_bindgen::from_value(request).unwrap();
+            let pre_req: IssuePreRequest = serde_wasm_bindgen::from_value(request).unwrap();
+            let media = match pre_req.meta {
+                Some(media) => {
+                    let media = crate::rgb::import_uda_data(media).await;
+                    Some(IssueMediaRequest::from(media.unwrap()))
+                }
+                None => None,
+            };
```
oacdutra: How we work when occurs error here?

oacdutra (changes requested):
Just naming convention adjustments.

Furthermore, shouldn't we change the name in web.rs, adding a "full_" prefix to the method, just like transfer?

oacdutra (approved):

PR #430, opened 2023-12-07
Support Auction Swaps

### LINKED ISSUE #365: replace Esplora::broadcast with direct request
When the error reported here: https://github.com/diba-io/bitmask-core/issues/364 and maybe in other occasions, all process are done but the transaction is not sent to the blockchain.

So, we need:

- [ ] verify_transfers if after some minutes the transactions is not in the mempool nor confirmed, we have to delete the transfer from the list
- [ ] list_contracts in AllocationDetail we need another property, onChain or inMempool, or something like that so if there are a transactions that we can't see yet, we don't take account this transaction

### LINKED ISSUE #413: Sequential Swaps
In BMC, we'll need to support 3 kinds of swaps so far:
- P2P swaps
- Auction swaps
- Sequential swaps

Only P2P swaps make sense with RGB20, the rest are for RGB21.

Due to how RGB works, if an issuer wants to sell multiple items in a collection, they'll need to create a new UTXO, and then assign the token change to a new witness UTXO. So, items can only be sold sequentially, not out of order, since the change needs to go to a new UTXO.

Later if a specific RGB21 asset is desired, they can be put up for sale by individuals in auction swaps or P2P swaps on their own UTXOs.

Multiple bids can be made on a sequential swap, and the highest will be chosen by the auction coordinator. Unfortunately we have no opportunity for Vickrey auctions due to the PSBTs for bids being signed in advance.

We'll need to be able to provide a Vec of IssueRequests. These will be encrypted to Carbonado by bitmaskd as offers, preferably in separate files so sold asset files can be removed. These are then used to satisfy bids by bitmaskd as a coordinator.

We will also need the capability to specify multiple fee recipients.

### LINKED ISSUE #422: Sequential Swaps L1 Transaction Batching
I believe it's possible to batch the sequential swaps PSBTs into a single transaction to save on fees. So, if the previous sequential swap transaction has not yet been mined, it might be possible to RBF the RGB21 change output and create a new transaction with another swap embedded. This requires keeping the old PSBTs around in case they're needed for tx batching. This should cut down on fee costs by up to 50%. This could then keep the fee savings on the RGB21 change output.

Depends #413

oacdutra:
**Description**
This PR intent supports sequential swaps and unifies the previous swap operations (now called "strategies").

Now, BMC have three ways to make swap operations:
1. **P2P:** The strategy allows participants (sellers and buyers) to make swaps without a coordinator. The match engine used is FIFO, i.e., the first bid that arrives with the correct target price.
2. **Airdrop:** The strategy allows participants (issuers) to make individual or collection airdrops. The airdrop respect the FIFO (First In First Out).
3. **Auction:** The strategy allows participants (sellers) to make individual or collection offers. The match engine used is the highest price, i.e., after the offer expires (or finished by seller), the coordinator selects the highest price bid and executes. ~Then, the coordinator updates the next available offer with the new UTXO (in case of the collections)~. All offers of the collections are publish and allow to receive bids.
4. **HotSwap:** The strategy allows participants (sellers and buyers) to make swaps without a coordinator. The main difference of the P2P is the sign condition. The sign condition is post-sign, i.e., the seller is the last to sign.

Also, It was necessary to carry out some refactorings to increase the maintainability of the swaps in BMC, according to the checklist below:

**Sequential Swaps (epic)**
- Choose Swap Strategies
  - [x] Add `HotSwap` instead of `presig=false` flag
  - [x] Add `P2P` instead of `presig=true` flag
  - [x] Add `Auction` to allow use coordinator
  - [x] Refactor swap tests
- Allow offer collections 
  - [x] Add `internal_create_buyer_bid` to allow reuse step
  - [x] Add `internal_create_seller_offer` to allow reuse step
  - [x] Allow vector of offers have the same `bundle_id`
  - [x] Add `create_auction_bid` to allow create bid (sequential swaps)  
  - [x] Add `create_auction_offers` to allow create one or more offers (collections in sequential swaps context)
  - [x] The `create_auction_offers` needs return all offers published in orderbook
- Now Public Orders is first citizen
  - [x] Allow change status **orders**  in public orderbook
  - [x] Use public orderbook to create **rgb transfers**
  - [x] Use public orderbook to update participants orders(aka. my orders)
- Match Engine based in swap strategies
  - [x] P2P is always FIFO
  - [x] Hotswap is always a seller decision
  - [x] Auctions is always highest bid and timebox.
  - [x] Airdrop is always FIFO
- Remove potential exploits from Public Orders information
  - [x] _[Airdrop]_ Prevent malicious buyers broadcast transactions before the **Auction** ends
  - [x] _[Auction]_ Prevent malicious buyers broadcast transactions before the **Auction** ends
  - [x] _[HotSwap]_ Prevent malicious seller broadcast multiple transactions with same UTXO (always force batch operation)
  - [x] _[P2P]_ Prevent malicious buyer broadcast multiple transactions with same UTXO (block use) 
  - [x] _[All]_ Storage `tapret commits` only DH (Diffie–Hellman) carbonado data
  - [x] Create integration tests
- Next Offer (aka.  Sequential Offer)
  - [x] Allow coordinator check and publish all transaction auctions (after offer finished)
  - [x] Allow seller finish the offers and launch new offers with remaining assets
  - [x] After confirmation of the previous offer transaction, update all subsequent offers with new UTXO
  - [x] Cancel all bids for offer already filled (executed)
  - [x] Prevents reuse of bitcoin utxos (force update dustless verification)
  - [x] Create integration tests
  - [x] Create integration tests (security)
- Expose endpoints for Bitmask Server
  - [x] Create/Remove Auction Offer
  - [x] Create/Remove Auction Bid
  - [x] Finish Auction Offers
  - [x] Renew Auction Offers
  - [x] Get Offer Collection 
  - [x] Get Bids by Offer Collection 
- Update Wasm API 
  - [x] Update Structs
  - [x] Update Methods
  - [x] Expose new Methods
  - [x] Create wasm tests for seq. swaps

Closes #413 
Closes #365 
Closes #422 

PS: I added #422 here because we can resolve all offers/bids in one "big" L1 transaction.

### DIFF
--- a/src/carbonado.rs
+++ b/src/carbonado.rs
@@ -414,12 +414,36 @@ mod client {
     }
 
     pub async fn auctions_store(
-        _bundle_id: &str,
-        _name: &str,
-        _input: &[u8],
+        bundle_id: &str,
+        name: &str,
+        input: &[u8],
         _metadata: Option<Vec<u8>>,
     ) -> Result<(), CarbonadoError> {
-        todo!()
+        let body = Arc::new(input.to_vec());
+        let network = NETWORK.read().await.to_string();
+        let endpoints = CARBONADO_ENDPOINT.read().await.to_string();
+        let endpoints: Vec<&str> = endpoints.split(',').collect();
+        let requests = Array::new();
+
+        for endpoint in endpoints {
+            let url = format!("{endpoint}/auction/{bundle_id}/{network}-{name}");
+            let fetch_fn = future_to_promise(fetch_post(url, body.clone()));
+            requests.push(&fetch_fn);
+        }
+
+        let results = JsFuture::from(Promise::all_settled(&JsValue::from(requests)))
+            .await
+            .map_err(js_to_error)?;
+
+        info!(format!("Store results: {results:?}"));
+
+        let results = serde_wasm_bindgen::from_value::<Vec<PostStorePromiseResult>>(results)?;
+        let success = results.iter().any(|result| result.value == 200.0);
+        if success {
+            Ok(())
+        } else {
+            Err(CarbonadoError::AllEndpointsFailed)
+        }
     }
 
     pub async fn retrieve_metadata(sk: &str, name: &str) -> Result<FileMetadata, CarbonadoError> {
@@ -535,10 +559,28 @@ mod client {
     }
 
     pub async fn auctions_retrieve(
-        _bundle_id: &str,
-        _name: &str,
+        bundle_id: &str,
+        name: &str,
     ) -> Result<(Vec<u8>, Option<Vec<u8>>), CarbonadoError> {
-        todo!()
+        let network = NETWORK.read().await.to_string();
+        let endpoints = CARBONADO_ENDPOINT.read().await.to_string();
+        let endpoints: Vec<&str> = endpoints.split(',').collect();
+
+        let requests = Array::new();
+        for endpoint in endpoints.iter() {
+            let url = format!("{endpoint}/auction/{bundle_id}/{network}-{name}");
+            let fetch_fn = future_to_promise(fetch_get_byte_array(url));
+            requests.push(&fetch_fn);
+        }
+
+        let result = JsFuture::from(Promise::any(&JsValue::from(requests)))
+            .await
+            .map_err(js_to_error)?;
+
+        let array = Uint8Array::from(result);
+        let encoded = array.to_vec();
+
+        Ok((encoded.to_vec(), None))
     }
 
     async fn fetch_post(url: String, body: Arc<Vec<u8>>) -> Result<JsValue, JsValue> {
--- a/src/carbonado.rs
+++ b/src/carbonado.rs
@@ -9,19 +9,24 @@ pub mod error;
 pub mod metrics;
 
 #[cfg(not(target_arch = "wasm32"))]
-pub use server::{handle_file, retrieve, retrieve_metadata, server_retrieve, server_store, store};
+pub use server::{
+    auctions_retrieve, auctions_store, handle_file, marketplace_retrieve, marketplace_store,
+    retrieve, retrieve_metadata, store,
+};
 
 #[cfg(not(target_arch = "wasm32"))]
 mod server {
-    use crate::constants::get_marketplace_nostr_key;
+    use crate::constants::{get_coordinator_nostr_key, get_marketplace_nostr_key};
 
     use super::*;
 
     use std::{
         io::{Error, ErrorKind},
         path::PathBuf,
+        str::FromStr,
     };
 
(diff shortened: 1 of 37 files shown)


josediegorobles:
Hi, could you update the wasm api?

oacdutra:
> Hi, could you update the wasm api?

Yes, I think I finishing it today.

oacdutra:
> Prevents reuse of bitcoin utxos (force update dustless verification)

Suppose we have a seller who wanted to sell 5 items and wants to sell via sequential swap. Throughout operations, the UTXO related to assets may decrease to smaller fractions until it turns to dust. As the seller no longer changes the PSBTs after the start of the auctions (the coordinator manages it), I thought about making a "minimum satoshis for operation" rule for the RGB20 and RGB21 wallets, what do you think?

src/carbonado.rs
```diff
@@ -78,6 +83,43 @@ mod server {
         Ok((filepath, body))
     }
 
+    pub async fn auctions_store(
```
cryptoquick, line 88: Is there any way to abstract this out so it's not in `src/carbonado.rs` methods? Ideally this file would be cleaner and simpler, more focused and have less duplication.

cryptoquick (changes requested):
Really good work on this! I like the clarification on some of the terminology, like server -> marketplace and presig -> strategy. Also, good catch on realizing we're using PSBTv0. As you might know, there's no PSBTv1... Only v0 and v2.

I just have one issue with code organization in carbonado.rs... It's really making that file quite a bit more complex than it should be. If I had to guess, that logic is being put in the wrong place. Do you agree?

oacdutra:
> > Prevents reuse of bitcoin utxos (force update dustless verification)
> 
> Suppose we have a seller who wanted to sell 5 items and wants to sell via sequential swap. Throughout operations, the UTXO related to assets may decrease to smaller fractions until it turns to dust. As the seller no longer changes the PSBTs after the start of the auctions (the coordinator manages it), I thought about making a "minimum satoshis for operation" rule for the RGB20 and RGB21 wallets, what do you think?

What do you guys think, @cryptoquick and @josediegorobles ?

src/carbonado.rs
oacdutra: I believe so, but I feel like it will lose some of its purpose. I centralized in this file all the calls that are sometimes made by the client and sometimes made by the server.

But yes, I can carry out another separation.

cryptoquick:
@crisdut We should always have an input from the buyer for paying fees and paying for the dust output.

Also, there's no way to batch transfers with auctions because they're separate PSBTs, correct? That's fine so long as multiple sequences can be spent within the same block.

josediegorobles (approved):
LGTM

PR #436, opened 2023-12-12
Update struts in typescript

josediegorobles:


lib/web/rgb.ts
```diff
@@ -793,10 +793,55 @@ export interface RgbOfferRequest {
...
+  fee_value: number,
+}
+
+export interface RgbSwapStatusResponse {
+    /// Transfer ID
+    consig_id: string,
+    /// Offer ID
+    offer_id: string,
+    /// Bid ID
+    bid_id: string,
```
cryptoquick: Let's camelCase these

cryptoquick (changes requested):

lib/web/rgb.ts
```diff
@@ -868,6 +868,8 @@ export interface RgbOfferResponse {
   sellerAddress: string;
   /// Seller PSBT (encoded in base64)
   sellerPsbt: string;
+  /// Bundle ID (collection)
+  bundleId: string[],
```
oacdutra:
```suggestion
  bundleId?: string,
```

PR #438, opened 2023-12-13
Fix multi-day live metrics

cryptoquick:
So, I noticed that at midnight, the Hut8 server resets our live metrics to 0, so we still need to collect metrics every night in order to tally the last day. That's the simplest way around the issue, at least. Still, this takes our server from running metrics 6 times a day to just once.

Also, I have ChatGPT help me with the solution to calculating minutes to midnight. Hopefully that code makes sense.


josediegorobles (approved):
LGTM.
Makes sense

PR #440, opened 2023-12-18
Use a database and JSON file for persistent, multi-day, live metrics.

cryptoquick:
This should be a superior approach that addresses a concern @crisdut raised about using a Lazy RwLock and also should lessen the burden on the server.

Supersedes #438

### DIFF
--- a/src/carbonado/metrics.rs
+++ b/src/carbonado/metrics.rs
@@ -1,18 +1,18 @@
 #![cfg(not(target_arch = "wasm32"))]
 use std::{
-    collections::{BTreeMap, BTreeSet},
+    collections::BTreeMap,
     env,
     path::{Path, PathBuf},
-    sync::Arc,
     time::SystemTime,
 };
 
 use anyhow::{anyhow, Result};
 use chrono::{DateTime, Duration, NaiveDate, Utc};
-use log::debug;
+use log::{debug, error};
 use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
-use tokio::{fs, sync::RwLock};
+use sled::{Config, Db, Mode};
+use tokio::fs;
 use walkdir::WalkDir;
 
 #[derive(Serialize, Deserialize, Default)]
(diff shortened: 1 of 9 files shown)


josediegorobles:
Lots of tests are failing

oacdutra:
> Lots of tests are failing

I will check this. The error being described is related to file manipulation.

src/carbonado/metrics.rs
```diff
@@ -287,9 +315,102 @@ pub async fn update(path: &Path) -> Result<()> {
...
+    } else {
+        METRICS_DB
+            .open_tree(DB_DAYS)?
+            .insert(day.as_bytes(), &[1])?;
+        true
+    };
+
+    let dir = env::var("CARBONADO_DIR").unwrap_or("/tmp/bitmaskd/carbonado".to_owned());
+    let mut metrics: MetricsData =
+        serde_json::from_str(&fs::read_to_string(format!("{dir}/metrics.json")).await?)?;
```
oacdutra, line 342: When we run automated tests, no integration test initializes these files, which ends up causing the error.

Could we add a check here to see if the file exists?

oacdutra (changes requested):

cryptoquick:
@crisdut Can you check if bitmaskd isn't being instantiated in time for it to ensure the directory is initialized for backend integration tests?

oacdutra:
> Can you check if bitmaskd isn't being instantiated in time for it to ensure the directory is initialized for backend integration tests?

We instance `bitmaskd` only for wasm32 tests. For integration tests, we call functions directly.

We need a verification like this:

```rust
    let dir = env::var("CARBONADO_DIR").unwrap_or("/tmp/bitmaskd/carbonado".to_owned());
    if !Path::new(&format!("{dir}/metrics.json")).exists()
        || !Path::new(&format!("{dir}/metrics.csv")).exists()
    {
        init(Path::new(&dir)).await?;
    }
```

oacdutra (approved):

josediegorobles (approved):
ACK

PR #442, opened 2023-12-21
Recompute keys from mnemonic whenever decrypting an existing wallet.

cryptoquick:
I've done some experimenting, and I've noticed that I sometimes on certain wallets get a different xpub. This recomputes the keys whenever decrypting the wallet, and after some testing, I've found those keys to be correct.


cryptoquick:
Never mind, this doesn't fix the problem, and it also breaks the paper wallet feature

PR #445, opened 2023-12-27
Fix Bump Fee

oacdutra:
**Description**

After investigating the error in  `create_bdk_rbf_transaction`, I noticed the strange behavior in the **BDK rbf** method. When we create a transaction builder from the `Wallet::build_fee_bump` method with **descriptor** and **change_descriptor**, the **BDK** gets all outputs recipients, including "final change output", and this avoids **BDK** apply replace by fee policy.

For example, 
Suppose you have 100.000 sats and create a new transaction, sending 10.000 sats to the address; the BDK generates the following transaction:

In   (1):   100.000
out (1):   10.000
out (2):   98.000

fee:        1.000

When we try to send a new transaction with **descriptor** and **change_descriptor**,  the BDK retrieves both outs instead of only the first out (10.000 sats). Then, BDK raises an inflation exception.

### DIFF
--- a/lib/web/bitcoin.ts
+++ b/lib/web/bitcoin.ts
@@ -95,9 +95,9 @@ export const drainWallet = async (
 export const bumpFee = async (
   txid: string,
   feeRate: number,
+  broadcast: boolean,
   descriptor: string,
-  changeDescriptor: string,
-  broadcast: boolean
+  changeDescriptor?: string,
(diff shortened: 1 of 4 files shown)


cryptoquick:
Interesting... Good to know.

cryptoquick:
Thank you for looking into this!

PR #449, opened 2024-02-03
fix: taproot derived xpub

jasonribble:
The example seed phrase was providing the the main extended public key. i.e derivation path `m` This can be confirmed on [bip32.net](https://bip32.net/)

Here's the `m` derivation in the current version:

image

image

We needed to provide an `xpub` that included the taproot descriptor. Since we already had the function `watcher_xpub`, that's all that needed to be switch out.

Now, if a user wants to copy the xpub for a watch only wallet; they will see

With the additional test, we can see that we now have the proper watch-only address.

### DIFF
--- a/src/bitcoin/keys.rs
+++ b/src/bitcoin/keys.rs
@@ -210,7 +210,7 @@ pub async fn get_mnemonic(
     };
 
     let public = PublicWalletData {
-        xpub: xpub.to_string(),
+        xpub: watcher_xpub.clone(),
         xpubkh,
         watcher_xpub,
         btc_descriptor_xpub,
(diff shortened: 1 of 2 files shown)


cryptoquick (approved):
ACK

Great work, nice and simple

jasonribble:
What do we do with the failing `Rust/test-wasm`?

oacdutra:
> What do we do with the failing `Rust/test-wasm`?

Apparently, the problem is related to the ehadless execution timeout, as all tests passed.

cryptoquick:
It always depends on why the test fails. If it's just an issue with CI, we're not too worried about it.

jasonribble:
How exciting! Thank you.

josediegorobles (approved):

